PR #6, opened 2021-01-26
Small bugfixes

blaze33:
`tesseractArray` is now 0 indexed but `G['giftDRPower']` isn't.
Just fixed indexes accordingly, should fix the time acceleration bug in particular (game was way faster than usual)

(sorry for the newline at end of file, editing via github added it without telling me first before committing...)

### DIFF
--- a/src/CheckVariables.ts
+++ b/src/CheckVariables.ts
@@ -274,6 +274,9 @@ export const checkVariablesOnLoad = (data: Player) => {
         c15RewardUpdate();
         player.loadedDec16Vers = true;
     }
+    if (data.talismanLevels[0] === null) {
+      player.talismanLevels = data.talismanLevels.slice(1);
+    }
 
     // in old versions of the game (pre 2.5.0), the import function will only work
     // if this variable = "YES!". Don't ask Platonic why.
--- a/src/CheckVariables.ts
+++ b/src/CheckVariables.ts
@@ -274,12 +274,6 @@ export const checkVariablesOnLoad = (data: Player) => {
         c15RewardUpdate();
         player.loadedDec16Vers = true;
     }
-    if (data.talismanLevels[0] === null) {
-      player.talismanLevels = data.talismanLevels.slice(1);
-    }
-    if (data.antUpgrades[0] === null) {
-      player.antUpgrades = data.antUpgrades.slice(1);
-    }
 
     // in old versions of the game (pre 2.5.0), the import function will only work
     // if this variable = "YES!". Don't ask Platonic why.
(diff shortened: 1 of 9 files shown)


Pseudonian:
Hello, ideally we'd like to have the index of `G['giftDRPower']` also be a zero-index (this may be found in variables.ts) as we are converting 1-index arrays to 0-index in the rewrite. I can merge the pull request when this is done and corresponding changes are made!

blaze33:
Ok, so I changed `G['giftDRPower']` and `G['giftbase']` to be also 0 indexed.

I added a fix for the bonus runes computation, one line was commented and the runes added by the talismans kept increasing without end.

While I'm at it, I noticed that `player.talismanLevels` was also converted to be 0 indexed, however when loading a pre 2.5 save, the data is not converted from the 1 indexed format. Is `CheckVariables.ts` the correct place to add such a check? Something like the following maybe:
```js
if (data.talismanLevels[0] === null) {
    player.talismanLevels = data.talismanLevels.slice(1)
}
```

src/Tax.ts
```diff
@@ -75,7 +75,7 @@ export const calculatetax = () => {
         exponent *= 0.5
     }
     G['maxexponent'] = Math.floor(275 / (Decimal.log(1.01, 10) * exponent)) - 1
-    const a2 = Math.min(G['maxexponent'], Math.floor(Decimal.log(G['produceTotal'].add(1).toNumber(), 10)));
+    const a2 = Math.min(G['maxexponent'], Math.floor(Decimal.log(G['produceTotal'].add(1), 10)));
```
KhafraDev, line 78: Pushed in a828f16b3e467bfb0fbd98d6252dcf7e48ee61ed, didn't see it here. My bad.

src/CheckVariables.ts
```diff
@@ -274,6 +274,9 @@ export const checkVariablesOnLoad = (data: Player) => {
         c15RewardUpdate();
         player.loadedDec16Vers = true;
     }
+    if (data.talismanLevels[0] === null) {
```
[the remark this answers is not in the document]
blaze33: Ah I hadn't see this piece of code, I now see it's because `talismanLevels` and `antUpgrades` are still defined with a `null` at index 0 in the player definition in Synergism.ts that the length check isn't triggered.
Just did a quick test and indeed just removing the nulls was sufficient, I'll remove those 2 unnecessary checks from the PR.

src/Tax.ts
blaze33: No worries :) Glad it's fixed!

KhafraDev:
@Pseudonian pull request LGTM. Once you do some testing it'll be good to merge.

KhafraDev (approved):

Pseudonian:
Looks good to merge - and it has! Thank you for the PR, we greatly appreciate it.

PR #13, opened 2021-01-30
Refactoring of roomba & talisman autobuyers for performance

kewne7768:
Currently, both the roomba and talisman autobuyers will end up calling `calculateRuneLevels()` and various other recalculations a couple of times each tick. This is extremely wasteful:
Screenshot_20210130_183725

This PR changes these so that the recalculations are only ran if a level is successfully bought, and if the roomba can't buy a research, it stops trying instead of failing an additional `player.challengecompletions[14]` times per second.

### DIFF
--- a/src/Research.ts
+++ b/src/Research.ts
@@ -13,9 +13,10 @@ const getResearchCost = (index: number, buyAmount = 1, linGrowth = 0): [number, 
  * Buys Research of index.
  * @param index 
  * @param auto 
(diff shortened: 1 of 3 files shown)


KhafraDev (approved):

Pseudonian (approved):

Pseudonian:
This is good to go! Thank you very much for the refactoring of these autobuyers; this has been merged!

PR #15, opened 2021-01-30
Webpack Dev Server

burnstreet:
This enables the webpack dev server via npm run start. you can access it at localhost:3001.
It has hot module reloading enabled, so if you change someting (I think only in ts code), it will automatically be applied without reloading the page.
in addition, I enabled source maps so in browser debugging is easier.

With the dev server, you could go and delete the dist folder from git and add it to .gitignore, that way you will get less conflicts there. normally one does not store generated code in git for that reason. If you do that, you will probably have to trigger npm run build on commit to still have the dist files on github.io, see https://docs.github.com/en/actions/guides/setting-up-continuous-integration-using-workflow-templates.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,2980 +1,8 @@
 {
   "name": "synergismofficial",
   "version": "2.5.0-alpha",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
-    "": {
-      "name": "synergismofficial",
-      "version": "2.5.0-alpha",
-      "license": "MIT",
-      "dependencies": {
-        "acorn": "^8.0.4",
-        "ajv": "^6.12.6",
-        "ajv-keywords": "^3.5.2",
-        "ansi-colors": "^4.1.1",
-        "ansi-styles": "^4.3.0",
-        "big.js": "^5.2.2",
-        "braces": "^3.0.2",
-        "break_infinity.js": "^1.2.0",
-        "browserslist": "^4.16.1",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -1,251 +1,99 @@
 {
   "name": "synergismofficial",
   "version": "2.5.0-alpha",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
KhafraDev, line 4: Would you mind updating your Nodejs version to >=15.0.0 so the package.json structure is updated to v2? Thanks!

webpack.config.js
```diff
@@ -1,20 +1,31 @@
...
         rules: [{
             test: /\.tsx?$/,
             use: 'ts-loader',
             exclude: /node_modules/,
         }],
+        
     },
     resolve: {
-        extensions: [ '.ts' ],
+        extensions: [ '.ts', '.js' ],
```
KhafraDev, line 16: no js files should be resolved

package-lock.json
burnstreet: I quite strongly prefer the LTS version of node / npm, as I use it in my day job.
nvm on windows has its issues, so I would prefer not having to update.
Do you use any of the feature node latest has over the LTS?

webpack.config.js
burnstreet: webpack dev server needs thatm otherwise it won't run.

package-lock.json
burnstreet: I nevertheless installed it once to spare you some work.

PR #19, opened 2021-02-03
Shop Description Fix

Xander374:
Consistency and may fix an undefined popping up there


KhafraDev:
The actual issue is that
```js
player.shopUpgrades.offeringAutoLevel 
```
was changed to
```js
player.shopUpgrades.offeringAuto
```
Due to the index signature in player.shopUpgrades that @Pseudonian didn't fix (remove), TypeScript will ignore all invalid keys and treat every key, even if it's not on the object, as a number.
https://github.com/Pseudo-Corp/SynergismOfficial/blob/f826ddd484ae1907f229dbb4ac97dd1f48c0c7ad/src/types/Synergism.d.ts#L350

Going to deny this as it doesn't solve the undefined, and it's a small subset of a much larger issue with  32bc87e.

PR #25, opened 2021-02-07
Fix display of talisman fragment inventory

jsherbach:
When fortifying, if levels go up by less than 30 then the displayed fragment inventory wouldn't update (good news is that the actual inventory updated appropriately -- it just wasn't being displayed).

Note: I'm a noob here, so I'll ask what might be a very dumb question. How do I start a test instance of the game to see if my fix does what I expect it to do? Any chance there's a guide someplace so I don't waste people's time answering questions like this?

### DIFF
--- a/src/Talismans.ts
+++ b/src/Talismans.ts
@@ -405,11 +405,16 @@ export const updateTalismanAppearance = (i: number) => {
     }
 }
 
+// Attempt to buy a fixed number of levels (number varies based on
+// ascension). Returns true if all levels were bought -- false otherwise
+// (whether due to lack of resources or hitting the level cap).
 export const buyTalismanLevels = (i: number, auto = false): boolean => {
     let max = 1;
     if (player.ascensionCount > 0) {
         max = 30
     }
+    let purchasedMax = true;
+    let hasPurchased = false;
     for (let j = 1; j <= max; j++) {
         let checkSum = 0;
         let priceMult = G['talismanLevelCostMultiplier'][i]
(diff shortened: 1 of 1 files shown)


kewne7768:
I don't think there's a guide yet but with the way Webpack is configured, it comes down to:

* Install Node.js
* Run `npm install` to grab Webpack and the game's libs
* Run `npm run start` to start a development server at http://localhost:3001

src/Talismans.ts
```diff
@@ -405,11 +405,16 @@ export const updateTalismanAppearance = (i: number) => {
     }
 }
 
+// Attempt to buy a fixed number of levels (number varies based on
+// ascension). Returns true if all levels were bought -- false otherwise
```
kewne7768, line 409: This should probably return true if any level was bought - it was supposed to represent any change to talisman shards, so that the autobuyer can update rune levels and inventory (but only once instead of for every talisman)

src/Talismans.ts
```diff
@@ -456,20 +461,21 @@ export const buyTalismanLevels = (i: number, auto = false): boolean => {
...
-    if (!auto) {
+    if (!auto && hasPurchased) {
         showTalismanPrices(i);
         // When adding game state recalculations, update the talisman autobuyer in tack() as well
         updateTalismanInventory();
         calculateRuneLevels();
     }
 
-    return true;
+    return purchasedMax;
```
kewne7768, line 478: Which would mean this would go to `hasPurchased` instead

src/Talismans.ts
jsherbach: Ah, rereading the autobuyer code that makes more sense -- for some reason i thought the autobuyer was running in a loop and only breaking out when this returns true (it is in a loop, but over the talis types :) and there's no breaking involved). Fixed

src/Talismans.ts
jsherbach: done

jsherbach:
Need to run -- will set up node and run a local test of this later today/tonight and confirm this actually fixes stuff then.

Pseudonian:
Hello, this is confirmed to work. Thank you for the contribution to v2.5.0!

Pseudonian (approved):

PR #30, opened 2021-02-09
added plugin to open cubes from console

Pomroka:
Plugin that exposes open cubes, teseract, hypercube and platonics to the global scope, so they can be use from console.


KhafraDev:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties

Once the 4 ``Object.defineProperty``s are converted I'll be happy to merge.

PR #32, opened 2021-02-10
Rework automation unlock code

kewne7768:
The automation unlocker is currently based on the order of elements with the `auto` class in the HTML. Recently, this class was added onto the Tesseract autobuyers, which broke a bunch of features: the toggles for the upgrade autobuyers, the unlock for upgrade hover-to-buy, and the confirmation buttons in the settings.

We could fix this by shuffling the elements around again, but instead I decided to fix it by hardcoding element IDs, as those change much less frequently than the order of the HTML (and when they change they're easy to find via Find in Files).

This should fix these reported autobuy issues: https://discord.com/channels/677271830838640680/730669598738743306/809056084810924083 aside from maybe the accel boost stuff, which doesn't seem to be a bug.

### DIFF
--- a/src/UpdateHTML.ts
+++ b/src/UpdateHTML.ts
@@ -286,50 +286,58 @@ export const revealStuff = () => {
 
     //I'll clean this up later. Note to 2019 Platonic: Fuck you
     // note to 2019 and 2020 Platonic, you're welcome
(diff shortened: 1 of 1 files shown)


KhafraDev:
... thanks cloudflare

PR #34, opened 2021-02-16
fix shard/fragment buy ammount

Pomroka:
Fix shards/fragment instead of buying 10/25/50/100 % was buying 0/1/2/3 %


Pseudonian:
Yes, this is good, thank you.

PR #35, opened 2021-02-24
some degree of css cleanup

omsi6:
- removed global scale transform
- centered general user interface on wide screens
- centered subtab buttons
- centered building rows
- added a hover effect to tabs and subtabs
- made your cursor only be a pointer on ant tiers and upgrades, building purchase buttons, and talisman shard purchase buttons if they're affordable
- added a hover effect to building purchase buttons and talisman shard purchase buttons when they're purchasable
- highlight the coin building and settings subtabs by default as they are the default subtabs on load
- cleaned up background css related to all these things (tried to use flexbox and classes rather than grids and html style decs)

### DIFF
--- a/Synergism.css
+++ b/Synergism.css
@@ -1,11 +1,9 @@
 html {
 	min-width: 1250px;
-	max-width: 1400px;
 	position: relative;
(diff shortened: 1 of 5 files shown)


Pseudonian (approved):

Pseudonian:
Looks good to merge. :D

PR #45, opened 2021-03-21
more css stuff

omsi6:
- fixed ant sacrifice settings buttons and the max toggle overlapping in certain edge cases
- added special hover class for active challenges
- remade the blessing and spirit tabs in flexbox and improved their centering
- made setting confirmation toggle icons slightly smaller to avoid them being blurry and better center them
- added hover effects and proper pointer cursor conditions to constant upgrade

### DIFF
--- a/Synergism.css
+++ b/Synergism.css
@@ -1368,6 +1368,9 @@ nav li:hover {
 .challengeActive {
 	background-color: plum;
 }
+.challengeActive:hover {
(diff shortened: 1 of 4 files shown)


Pseudonian (approved):

Pseudonian:
Looking good! Thank you for these CSS changes 💯

PR #48, opened 2021-04-24
yet more css stuff

omsi6:
- fixed the settings subtab version text pushing over all other elements on screens with a large horizontal width
- redid the shop tab in flexbox and changed the layout slightly
- added hover effects to the hepteract forge buttons
- fixed the respec all talismans button overlapping the talisman upgrade cost div when you only have a few talismans
- fixed manual auto research still visually applying the auto research class to researches if you tried to click a study while you have the roomba unlocked and enabled

### DIFF
--- a/Synergism.css
+++ b/Synergism.css
@@ -1043,6 +1043,7 @@ nav li:hover {
 
 .talismansContainer {
 	display: flex;
+	position: relative;
 	width: 70%;
     justify-content: space-around;
     padding: 0 10px;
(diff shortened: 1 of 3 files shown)


Synergism.css
```diff
@@ -1514,32 +1521,46 @@ nav li:hover {
...
+
+#actualShop > div {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    margin-top: 5px;
+    margin-bottom: 5px;
+    margin-left: 5px;
+    width: 100px;
+    white-space: nowrap;
```
KhafraDev, line 1563:
```suggestion
```
Adding this causes the text to not wrap when it overflows:
image

KhafraDev (changes requested):

Synergism.css
omsi6: I've made this now apply only to the level displays rather than the purchase buttons, as that was the intended change

KhafraDev (approved):

PR #51, opened 2021-06-19
Clicking a contributor now opens a new tab

earthernsence:
Very small change but bugged me as soon as I clicked on a few names.


KhafraDev:
Fixed in e531db4f409e3d643bd3620c905c9bf851c97b3a along with some other changes, thanks.

PR #52, opened 2021-06-20
Added dashboard load to main loader

tomhughes-invento:
The button from test to load the dashboard was removed alongside the built-in logger, this change adds the dashboard load to the main game load so it is available right away


KhafraDev (changes requested):
-1

The plugin was removed when we switched from Webpack to esbuild since esbuild doesn't support dynamic imports from IIFEs. Adding it back would increase the bundle size (which is already quite large) and it isn't a decent solution to fix plugins.

I'll leave it up to @Pseudonian though.

tomhughes-invento:
Gonna close off this PR for now, and look into integrating the dashboard itself into S4N at a later date

PR #58, opened 2021-06-25
hide ant auto sac msg until relevant

hottestPotato:
I'm not sure if there are more places that this needs to go or not, but this is the least invasive way I can think of fixing this.


zachSummers:
The `autoAntSacrifice` span shows before sacrificing is possible, as well as before automated sacrificing is possible. I'm not steeped in how the code writes out the text to the span, but this seems to be how all the other UI elements are hidden until ready, so I added this here.

Pseudonian (approved):

PR #60, opened 2021-06-25
Corruption on c11

PotatoArt:
Many players get confused about corruptions the moment they unlock them by beating c10, even before they've ascended. Corruptions aren't very useful until tesseracts are unlocked and the player is strong enough to ascend with reasonable speed, both of which occur roughly when c11 is completed (or possible to complete). There isn't any other mechanic unlocked by c11, so this also serves to make introduction to new mechanics more gradual. 

The main downside of this change is that some players might have added corruptions but not have completed c11 yet. One way around the problem would be showing the corruption interface to any player with corruptions, another other option is cleansing anyone who hasn't completed c11, which is what's implemented here. 

Another downside is the lack of ability to automate ascends if the tab isn't visible containing the automation isn't visible. There are a few solutions to this: moving the automation to the cubes tab, not allowing automation to be purchased until after corruptions is unlocked, etc. I haven't implemented any of these things, but I can work whichever direction you'd prefer.

Also included is showing the ascension stats bar once the player ascends, rather than waiting until c11 is completed. I don't see why that information shouldn't be available to players who have yet to complete c11. (though, Tess has to be hidden until tess can be earned.) If this should be split into a separate pull request, I'm happy to do that.

### DIFF
--- a/index.html
+++ b/index.html
@@ -209,7 +209,7 @@
                 <li class="reincarnationunlock" id="researchtab">Research</li>
                 <li class="chal8" id="anttab">Anthill</li>
                 <li class="chal10" id="cubetab">WOW! Cubes</li>
-                <li class="chal10" id="traitstab">Corruption</li>
+                <li class="chal11" id="traitstab">Corruption</li>
                 <li id="settingstab">Settings</li>
                 <li class="reincarnationunlock" id="shoptab">Shop</li>
             </ul>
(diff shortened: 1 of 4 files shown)


Pseudonian:
This is great, but I think I will make it open for more testing for v2.6

PotatoArt:
Sounds good

PR #64, opened 2021-06-27
Updating challenge displays on all resets

PotatoArt:
Challenges display previously unlocked levels on reincarnation and ascension, even though their progress gets reset. Adding a call to `updateChallengeDisplay()` to all resets should solve the problem.


PotatoArt:
Ok, I'm real bad at managing branches apparently, so now this PR also contains an update to Platonic upgrade text for Alpha and Beta

Pseudonian (approved):

PR #66, opened 2021-06-27
Fixing achievement effect

PotatoArt:
This includes changes from another PR #64 since I'm really bad at branch management on my silly fork, but oh well. I'm a pro I swear, but the stakes are so low on my own fork that I got lazy about it.

The description of achievement 254 [found here](https://github.com/Pseudo-Corp/SynergismOfficial/blob/master/src/Achievements.ts#L488) says it's a 10-15% boost, but the actual multiplier is 2-4x, which doesn't seem correct. Either the description is off, or the implementation is, so here I've set the implementation to match the description, albeit without the same rounding.


Pseudonian:
That is correct... THank you for the edits :D

PR #68, opened 2021-06-27
Tess autobuy fix

PotatoArt:
Another PR with code from #64 . I'll wait for that one to be merged or rejected before making more now. 

Instead of only checking if buying a single building would save enough tess, check if buying `tesseractbuyamount` would save enough tess.


PotatoArt:
This PR doesn't seem to have actually been effective.

mcpower:
This PR has a slight bug with the calculation of buying multiple tesseract buildings (you need the sum of each building's cost instead of just the final cost). There's two ways of fixing it:

- calculate the correct cost to buy multiple, or
- always autobuy one at a time instead.

I've implemented the second one in the above mentioned PR, which also fixes a correctness problem with buying multiple tesseract buildings at a time.

PR #69, opened 2021-06-28
Adding alt text to some things

PotatoArt:
Added alt text to all the resource icons and reset buttons, and all the Upgrade tab upgrades. Hopefully this helps people find the reset buttons and upgrades easier.

Also added a hopefully-helpful dynamic adding and removing of `aria-describedby` to things when hovered, so a screen reader will automatically read the dynamically updated descriptions fields right after reading the image's alt text. 

Keeping the score small for this, want to get this into the hands of someone who actually know what they're doing with screen readers and keyboard navigation to tell if this is actually helpful or not before going all in on everything else. My screen reader does what I expect it to when hovering around in a jsfiddle, but I have no idea how actual users of keyboard navigation do keyboard navigation.

### DIFF
--- a/index.html
+++ b/index.html
@@ -136,59 +136,59 @@
             <div class="currenciesContainer">
                 <div class="currencyRow">
                     <div class="currencyContainer">
-                        <img class="img coin" src="Pictures/Coin.png" title='Coin' alt="" loading="lazy">
+                        <img class="img coin" src="Pictures/Coin.png" title='Coin' alt="Coin" loading="lazy">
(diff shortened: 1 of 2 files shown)


PotatoArt:
This should probably get serious testing if possible, it's really easy for attempts at making things accessible to make things much worse. (That's about the extent of my knowledge about accessibility.) Plus, I'm pretty sure there are more than 0 typos.

Pseudonian:
This will need to be tested, I could make this particular for v2.6 or next major update. I do appreciate the contribution though!

PR #70, opened 2021-06-28
Autobuy all tesseract buildings simultaneously

mcpower:
This fixes the bug mentioned in #68 (by not buying multiples at all), and fixes some edge cases with the previous logic.

As a counterexample to the previous logic, imagine:

- No buildings bought (costs are [1, 10, 100, ...])
- tesseractbuyamount is 10.
- Player has 100 tesseracts.

The previous logic would buy two tier 2s (10+80=90), which is cheaper than buying four tier 1s (1+8+27+64=100) or one tier 3 (100). However, the expected behavior here is to buy three tier 1s and one tier 2.

~~As a result, however, autobuying is slower and takes N ticks to buy N buildings. Let me know if this is a problem - I can implement an O(log tesseracts) binary search to buy the exact amount of buildings needed instead of buying them one-by-one if needed.~~

Instead of buying one-by-one, this PR now buys all tesseracts in one shot, even with ridiculous numbers of tesseracts (accurate up to 1e47 tesseracts, still works up to 1e300).

### DIFF
--- a/src/Buy.ts
+++ b/src/Buy.ts
@@ -719,7 +719,181 @@ export const tesseractBuildingCosts = [1, 10, 100, 1000, 10000] as const;
 //   tesseractBuildingCosts[i-1] * n^3.
 // so the first n tesseract buildings of tier i costs
 //   cost(n) = tesseractBuildingCosts[i-1] * (n * (n+1) / 2)^2
-// in total. Use cost(owned+buyAmount) - cost(owned) to figure the cost of 
+// in total. Use cost(owned+buyAmount) - cost(owned) to figure the cost of
+// buying multiple buildings.
+
+export type TesseractBuildings = [number|null, number|null, number|null, number|null, number|null];
+
+const buyTessBuildingsToCheapestPrice = (ownedBuildings: TesseractBuildings, cheapestPrice: number): [number, TesseractBuildings] => {
+    const buyToBuildings = ownedBuildings.map((currentlyOwned, index) => {
+        if (currentlyOwned === null) {
+            return null;
+        }
+        // thisPrice >= cheapestPrice = tesseractBuildingCosts[index] * (buyTo+1)^3
+        // buyTo = cuberoot(cheapestPrice / tesseractBuildingCosts[index]) - 1
+        // If buyTo has a fractional part, we want to round UP so that this
+        // price costs more than the cheapest price.
+        // If buyTo doesn't have a fractional part, thisPrice = cheapestPrice.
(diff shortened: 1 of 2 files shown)


mcpower:
Also note that `accelerate` was implemented in `buyTesseractBuilding` before this commit - presumably because autobuying was too slow before?

PotatoArt:
Buying one building at a time will mean it will take a dozen seconds to spend a full day's worth of tesseracts. This is potentially annoying, tesseract building production depends on time acceleration to be productive, so players need to sit in an ascend without Spatial Dilation for additional time. Before `accelerate` was applied, buying a day's worth of buildings 1 at a time would typically take two or three minutes. I wouldn't call a few seconds a problem, but multiple minutes would be annoying.

Also, thanks for taking a look at this.

mcpower:
Thanks for that information @PotatoArt! I'll implement something which will take one tick to buy everything at once, so there's no need to wait.

mcpower:
This is now implemented! Do note that **THIS HAS NOT BEEN TESTED IN-GAME**, and I have not tested this apart from a few calls to `calculateTessBuildingsInBudget`. Please test this in-game if you do want to merge this!

This is inexact starting from 1e47 tesseracts due to exceeding `Number.MAX_SAFE_INTEGER` buildings, but I assume that will never happen.

mcpower:
I've tested this in game and it works as expected.

mcpower:
@KhafraDev Ping - this change is ready to be reviewed. This should also fix the autobuyer being buggy with if `tesseractbuyamount` isn't set to 1.

src/Buy.ts
```diff
@@ -715,31 +715,219 @@ export const buyParticleBuilding = (
...
+        actualBuy = buyFrom + amount;
     }
-    const actualCost = intCost * Math.pow(actualBuy * (actualBuy + 1) / 2, 2) - subCost
+    const actualCost = intCost * Math.pow(actualBuy * (actualBuy + 1) / 2, 2) - subCost;
     return [actualBuy, actualCost];
 }
 
-export const buyTesseractBuilding = (index: OneToFive, accelerate = false) => {
+export const buyTesseractBuilding = (index: OneToFive, amount?: number) => {
+    amount ??= player.tesseractbuyamount;
```
KhafraDev, line 922:
```suggestion
export const buyTesseractBuilding = (index: OneToFive, amount = player.tesseractbuyamount) => {
```

KhafraDev (changes requested):
this is the only issue I saw, Pseudo will test it eventually

Pseudonian:
I did not notice anything wrong while testing. This seems to work fine, so I will merge it for now. Thank you!

KhafraDev:
thanks, cloudflare...

PR #73, opened 2021-06-30
Speed up autochallenge

mcpower:
A common strat used for C15 involves 0/0/0 autochallenge to build up Obtainum. A recent PR (#64) slowed this down a bit, so this PR makes autochallenges use the fast path of `reset()` to avoid this slowdown.

### DIFF
--- a/src/Toggles.ts
+++ b/src/Toggles.ts
@@ -83,20 +83,20 @@ export const toggleChallenges = (i: number, auto = false) => {
     if (player.currentChallenge.transcension === 0 && (i <= 5)) {
         if(player.currentChallenge.ascension !== 15 || player.ascensionCounter >= 2){
             player.currentChallenge.transcension = i;
-            reset("transcensionChallenge", false, "enterChallenge");
+            reset("transcensionChallenge", auto, "enterChallenge");
             player.transcendCount -= 1;
         }
     }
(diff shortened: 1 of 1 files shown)


PotatoArt:
If this is specifically for 0/0/0 obtainium, then why do c1-5 and c11+ need to use the fast path? IIRC the 0/0/0 autochallenge obt grind turned off c1-5 from being automatically started, so that auto-challenge didn't waste ticks on challenges that didn't generate Obtainium.

Admittedly, there's no reason why Transcends/Transcend challenged would need to `revealStuff` or `updateChallengeDisplay`, and I don't think Ascend challenges can even by started by autochallenge, so it's a moot point really.

mcpower:
@PotatoArt All "auto" resets (auto prestige, auto transcension, auto ascension) already use the fast path, so I assume the fast path is intended for all resets not controlled by the user. Khafra / Platonic might know more though.

@KhafraDev Ping - this change is ready to be reviewed.

Pseudonian:
Needs a little more testing before I merge.

mcpower:
I don't believe this is safe to merge in - applying the same "optimisation" to exiting challenges causes graphical errors.

Closing this PR - feel free to reopen if you think this is safe.

PR #76, opened 2021-07-02
ant auto-sac hiding from v2.6

PotatoArt:
#58 but this time it's on master so we don't have to wait for 2.6 to get this in


Pseudonian:
Looks good to me.

PR #81, opened 2021-07-07
small fixes

ArchaDust:
Line 1792: upgrade 34 and 35 are meant to be applied additively to each other according to the weird formula in upgrade 35, I've formatted it in a way that also demonstrate the additive behavior

line 2017: the achievement for enabling Prism rune is 44, but the code uses 43, which makes Prism effect enabled early

line 2176: .times(dt / 0.025) can be applied once for everything


src/Synergism.ts
```diff
@@ -2014,7 +2009,7 @@ export const multipliers = (): void => {
     if (player.achievements[37] > 0.5 && player.prestigePoints.gte(10)) {
         G['globalCrystalMultiplier'] = G['globalCrystalMultiplier'].times(Decimal.log(player.prestigePoints.add(1), 10))
     }
-    if (player.achievements[43] > 0.5) {
+    if (player.achievements[44] > 0.5) {
```
KhafraDev, line 2012: is there any indication that player.achievements is 1-based? I couldn't find any in the code

ArchaDust: I have been extensively checking a lot of things and achievements were always 1 based from what I have seen, an easy way to check this is the Recycle calculation code, a ot of achievements are used there

ArchaDust:
Achievement 43/44 re: Prism is kinda huge early and might actually be a requirement to reach 1e6 mythos in a reasonable time, fixing this might make the "1 to 1e6 mythos" part harder than needed. I haven't tested this myself but might do during the day.
So the alternative solution would be to move the "unlock prism rune" part of 44's reward to 43's desc

src/Synergism.ts
```diff
@@ -1789,12 +1789,7 @@ export const updateAllMultiplier = (): void => {
     a *= (1 + player.achievements[58] / 100)
     a *= (1 + player.achievements[59] / 100)
     a *= Math.pow(1.01, player.upgrades[21] + player.upgrades[22] + player.upgrades[23] + player.upgrades[24] + player.upgrades[25])
-    if (player.upgrades[34] > 0.5) {
-        a *= 1.03 * 100 / 100
-    }
-    if (player.upgrades[35] > 0.5) {
-        a *= 1.05 / 1.03 * 100 / 100
-    }
+	a *= (1 + 1.03 * player.upgrades[34] + 1.02 * player.upgrades[35])
```
Pseudonian, line 1792: This rewrite is not correct, as this would give a 3.05x multiplier when both upgrades are bought, but intended is 1.05x.

src/Synergism.ts
```diff
@@ -2173,12 +2168,12 @@ export const resourceGain = (dt: number): void => {
...
-        player.coinsThisPrestige = player.coinsThisPrestige.add(addcoin.times(dt / 0.025));
-        player.coinsThisTranscension = player.coinsThisTranscension.add(addcoin.times(dt / 0.025));
-        player.coinsThisReincarnation = player.coinsThisReincarnation.add(addcoin.times(dt / 0.025));
-        player.coinsTotal = player.coinsTotal.add(addcoin.times(dt / 0.025))
+        const addcoin = (Decimal.min(G['produceTotal'].dividedBy(G['taxdivisor']), Decimal.pow(10, G['maxexponent'] - Decimal.log(G['taxdivisorcheck'], 10)))).times(dt / 0.025)
+        player.coins = player.coins.add(addcoin);
+        player.coinsThisPrestige = player.coinsThisPrestige.add(addcoin);
+        player.coinsThisTranscension = player.coinsThisTranscension.add(addcoin);
+        player.coinsThisReincarnation = player.coinsThisReincarnation.add(addcoin);
+        player.coinsTotal = player.coinsTotal.add(addcoin)
```
Pseudonian, line 2176: I will add this is a solid optimization choice.

Pseudonian (approved):

PR #83, opened 2021-07-10
Maintain last updated date automatically

tomhughes-invento:
Changes do a few things:

- Moves the version number & testing flag into a config file, and adds a new config for last updated date (stored with a placeholder for easy identification).
- Updates all places that used the version/testing vars from Synergism.ts to read them from Config.ts.
- Adds a node script that goes to the new config file and replaces the placeholder with a UTC definition for the current date.
- Adds a script to the package to run it.
- Updates the gh-pages workflow to run the script before it builds.
- Lint fixes from previous commit by other contributers.

Since version number isn't something that is generally maintained, this aims to at least keep a track of when a build was last pushed out to players.

### DIFF
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -1,3 +1,3 @@
 export const version = '2.5.5';
 export const testing = false;
-export const lastUpdated: Date = new Date('##LAST_UPDATED##');
+export const lastUpdated = new Date('##LAST_UPDATED##');
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -1,0 +1,3 @@
+export const version = '2.5.5';
+export const testing = false;
+export const lastUpdated: Date = new Date('##LAST_UPDATED##');
--- a/update-last-updated.js
+++ b/update-last-updated.js
@@ -1,0 +1,28 @@
+const {readFileSync, writeFileSync} = require('fs');
+
+const currentDate = new Date();
+const UTC = `Date.UTC(${currentDate.getUTCFullYear()}, ${currentDate.getUTCMonth()}, ${currentDate.getUTCDate()}, ${currentDate.getUTCHours()}, ${currentDate.getUTCMinutes()}, ${currentDate.getUTCSeconds()})`;
+console.log('\x1b[33m%s\x1b[0m', `Updating last updated config date to ${UTC}`);
+
+
+try {
+    console.log('\x1b[33m%s\x1b[0m', 'Reading file & updating ##LAST_UPDATED##')
+    let config = readFileSync(`${__dirname}/src/Config.ts`).toString();
+
+    if (!config.match(/'##LAST_UPDATED##'/g)) {
+        throw new Error('File does not contain the \'##LAST_UPDATED##\' placeholder');
+    }
+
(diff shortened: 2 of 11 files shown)


update-last-updated.js
```diff
@@ -0,0 +1,28 @@
+const {readFileSync, writeFileSync} = require('fs');
+
+const currentDate = new Date();
+const UTC = `Date.UTC(${currentDate.getUTCFullYear()}, ${currentDate.getUTCMonth()}, ${currentDate.getUTCDate()}, ${currentDate.getUTCHours()}, ${currentDate.getUTCMinutes()}, ${currentDate.getUTCSeconds()})`;
+console.log('\x1b[33m%s\x1b[0m', `Updating last updated config date to ${UTC}`);
+
+
+try {
+    console.log('\x1b[33m%s\x1b[0m', 'Reading file & updating ##LAST_UPDATED##')
+    let config = readFileSync(`${__dirname}/src/Config.ts`).toString();
```
KhafraDev, line 10:
```suggestion
    let config = readFileSync(`${__dirname}/src/Config.ts`, 'utf-8');
```
so we don't have to convert buffer -> string

update-last-updated.js
```diff
@@ -0,0 +1,28 @@
...
+
+    if (!config.match(/'##LAST_UPDATED##'/g)) {
```
KhafraDev, line 12:
```suggestion
    if (!/'##LAST_UPDATED##'/g.test(config)) {
```

update-last-updated.js
```diff
@@ -0,0 +1,28 @@
...
+        throw new Error('File does not contain the \'##LAST_UPDATED##\' placeholder');
+    }
+
+    config = config.replace(/'##LAST_UPDATED##'/g, UTC);
+
+    if (!config.match(/Date\.UTC\(/g)) {
```
KhafraDev, line 18:
```suggestion
    if (!/Date\.UTC\(/g.test(config)) {
```

update-last-updated.js
```diff
@@ -0,0 +1,28 @@
...
+        throw new Error('Couldn\'t replace the placeholder');
+    }
+
+    console.log('\x1b[33m%s\x1b[0m', 'writing file back to source')
+    writeFileSync(`${__dirname}/src/Config.ts`, Buffer.from(config));
```
KhafraDev, line 23:
```suggestion
    writeFileSync(`${__dirname}/src/Config.ts`, config);
```
doesn't need to be a Buffer

KhafraDev (changes requested):
I'd eventually like to add some env to CI so that we can differentiate people testing/working on PRs from a CI run but it seems alright for now.

src/Config.ts
```diff
@@ -0,0 +1,3 @@
+export const version = '2.5.5';
+export const testing = false;
+export const lastUpdated: Date = new Date('##LAST_UPDATED##');
```
KhafraDev, line 3:
```suggestion
export const lastUpdated = new Date('##LAST_UPDATED##');
```
we don't need this extraneous cast

KhafraDev (approved):

PR #86, opened 2021-07-12
fix import/load glitches & adjustments to offline ants accuracy

PlatypusKing:
Load/Import - There was a weird bug where loads and imports would not be consistent (even if done back to back, and wildly inconsistent if moving between different saves). I resolved this by taking calculateOffline out of loadSynergy and instead having it run after loadSynergy has completed as part of reloadShit.

Offline Ants - I adjusted calculateOffline so that all resource generating/impacting items are in the loop that simulates growth which has made the results completely accurate when compared to a game left open (from my testing at least).

### DIFF
--- a/src/Calculate.ts
+++ b/src/Calculate.ts
@@ -811,8 +811,7 @@ export const calculateOffline = (forceTime = 0) => {
     document.getElementById('preload').style.display = (forceTime > 0) ? 'none' : 'block';
     document.getElementById("offlineContainer").style.display = "flex";
 
-    player.offlinetick = (player.offlinetick < 1.5e12) ? (Date.now()) : player.offlinetick;
-    const runOffline = interval(resourceSimulator, 0)
+    player.offlinetick = (player.offlinetick < 1.5e12) ? (Date.now()) : player.offlinetick;    
 
     //Set the preload as a blank black background for now (to allow aesthetic offline counter things)
     const preloadImage = getElementById<HTMLImageElement>("preload"); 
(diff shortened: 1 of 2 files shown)


src/Calculate.ts
```diff
@@ -839,50 +840,48 @@ export const calculateOffline = (forceTime = 0) => {
...
     player.reincarnationCount += resetAdd.reincarnation;
-
+    
     timerAdd.ascension = player.ascensionCounter - timerAdd.ascension
     timerAdd.quarks = quarkHandler().gain - timerAdd.quarks
 
-    document.getElementById('offlineAscensionTimerNumber').textContent = format(timerAdd.ascension, 2, true)
-    document.getElementById('offlineQuarkCountNumber').textContent = format(timerAdd.quarks, 0, true)
+    //200 simulated all ticks [June 18, 2021]
+    function resourceSimulator() {
```
KhafraDev, line 854: please move this function into the interval above & make it an arrow function for consistency

KhafraDev (changes requested):

src/Calculate.ts
PlatypusKing: Not very familiar with arrow functions (I had just modified the function that already existed), but hopefully this is correct (it seems to run without issue, but want it to be consistent).  

If there is a better way to do this let me know and I'll make the adjustments.

```js
const runOffline = interval(() => resourceSimulator(), 0);
//200 simulated all ticks [July 12, 2021]
const resourceSimulator = () => {
...
```

KhafraDev: since resourceSimulator isn't used anywhere else in the code, it could be simplified to something like

```js
const runOffline = interval(() => {
    // do resourceSimulator stuff here
}, 0);
```

PlatypusKing: Ahhh makes that sense, sorry for the trouble.

Made the suggested change, thanks for the feedback/help Khafra

Pseudonian: quarks please

KhafraDev (approved):

Pseudonian (approved):

PR #94, opened 2021-07-17
KeyboardEvent.code for numeric events

### LINKED ISSUE #93: Use KeyboardEvent.code instead of key for intl keyboard layouts
Some keyboard layouts, most notably the French AZERTY, require the use of shift/caps lock to input numbers:
AZERTY

That means that without shifting, the QWERTY layout `1` key is `KeyboardEvent.key=1`, and on AZERTY is `KeyboardEvent.key=&`. It's kind of a pain when cycling through challenges, as you need to shift or caps lock to `1E2E` etc.

This can be remediated by using [`KeyboardEvent.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) instead, which uses physical key positions on a QWERTY keyboard regardless of the current layout. Thus, `key=1` becomes `code=Digit1`, or `code=Numpad1` on the num pad.

nmussy:
Fixes #93

### DIFF
--- a/src/Synergism.ts
+++ b/src/Synergism.ts
@@ -3260,7 +3260,7 @@ document.addEventListener('keydown', (event) => {
 
     const type = types[G['buildingSubTab']];
 
-    const key = event.key.toUpperCase();
+    const key = event.code.replace(/\D/g, '').toUpperCase();
     switch (key) {
         case "1":
         case "2":
--- a/src/Synergism.ts
+++ b/src/Synergism.ts
@@ -3260,7 +3260,7 @@ document.addEventListener('keydown', (event) => {
 
     const type = types[G['buildingSubTab']];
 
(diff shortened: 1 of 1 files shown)


KhafraDev (dismissed):
If this only applies to codes that start with `Digit` or `Numpad`, the regex should only be used on those codes.

```js
const key = event.code.startsWith('Digit') || event.code.startsWith('Numpad')
    ? event.code.replace(/\D/g, '').toUpperCase()
    : event.code.toUpperCase();
```

Without this, it will not work when the rest of these keybinds are eventually brought over to the reassignable hotkey area.

nmussy:
You got it, I've simplified your test with a single replace 👍

KhafraDev (approved):

PR #97, opened 2021-07-20
Improve the rendering of the individual stats buttons on the Stats fo…

darkscott:
…r Nerds tab.

image


darkscott:
I noticed the buttons were overlapping a lot. It might be fixed in someone else's work, but I got this to a reasonably pleasing arrangement...so if it hasn't been fixed yet elsewhere, you're welcome to take it :D

PR #107, opened 2021-07-23
fix max shop upgrade confirmation

### LINKED ISSUE #106: "Buy Max" shop upgrade doesn't affect the confirmation message
The new "Buy Max" option in the quark shop doesn't change the confirmation prompt. The user doesn't know how much quarks they're about to spend, or how many levels they'll gain

Before:
image

After:

image

nmussy:
Refactor `buyShopUpgrades` to simulate Buy Max purchase before prompting the user

Fixes #106

### DIFF
--- a/src/Shop.ts
+++ b/src/Shop.ts
@@ -360,21 +360,28 @@ export const shopDescriptions = (input: ShopUpgradeNames) => {
 
 export const buyShopUpgrades = async (input: ShopUpgradeNames) => {
     let p = true;
-    if (G['shopConfirmation']) {
-        p = await Confirm("Are you sure you'd like to purchase " + input + " for " + format(getShopCosts(input)) + " Quarks? Press 'OK' to finalize purchase.");
-    }
+    let upgradeCounter = 0;
+    let totalCost = 0;
 
-    if (p) {
-        if (G['shopBuyMax']) {
-            while (Number(player.worlds) >= getShopCosts(input) && player.shopUpgrades[input] < shopData[input].maxLevel) {
-                player.worlds.sub(getShopCosts(input));
-                player.shopUpgrades[input] += 1
-            }
-        } else {
-            if (Number(player.worlds) >= getShopCosts(input) && player.shopUpgrades[input] < shopData[input].maxLevel) {
-                player.worlds.sub(getShopCosts(input));
-                player.shopUpgrades[input] += 1
-            }
+    // simulate the total upgrade cost by iterating over the upgrade levels until we run out of quarks
+    do {
+        const cost = getShopCosts(input);
+        totalCost += cost;
+        player.worlds.sub(cost);
+        upgradeCounter++;
+        player.shopUpgrades[input]++;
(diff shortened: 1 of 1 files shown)


src/Shop.ts
```diff
@@ -360,21 +360,28 @@ export const shopDescriptions = (input: ShopUpgradeNames) => {
...
+        upgradeCounter++;
+        player.shopUpgrades[input]++;
+    } while(G['shopBuyMax'] && Number(player.worlds) >= getShopCosts(input) && player.shopUpgrades[input] < shopData[input].maxLevel);
+
     if (G['shopConfirmation']) {
-        p = await Confirm("Are you sure you'd like to purchase " + input + " for " + format(getShopCosts(input)) + " Quarks? Press 'OK' to finalize purchase.");
-    }
+        // revert the simulation by adding back the quarks and removing the shopUpgrades
+        player.worlds.add(totalCost);
+        player.shopUpgrades[input] -= upgradeCounter;
```
nmussy, line 378: I thought about only reverting if `!p` after the `Confirm`, but the user would see the decreased quarks total in the header etc. before confirming

src/Shop.ts
```diff
@@ -360,21 +360,28 @@ export const shopDescriptions = (input: ShopUpgradeNames) => {
 
 export const buyShopUpgrades = async (input: ShopUpgradeNames) => {
     let p = true;
+    let upgradeCounter = 0;
+    let totalCost = 0;
+
+    // simulate the total upgrade cost by iterating over the upgrade levels until we run out of quarks
+    do {
```
KhafraDev, line 367: it's not everyday that I see a do while loop!

src/Shop.ts
```diff
@@ -360,21 +360,28 @@ export const shopDescriptions = (input: ShopUpgradeNames) => {
...
-            while (Number(player.worlds) >= getShopCosts(input) && player.shopUpgrades[input] < shopData[input].maxLevel) {
-                player.worlds.sub(getShopCosts(input));
-                player.shopUpgrades[input] += 1
-            }
-        } else {
-            if (Number(player.worlds) >= getShopCosts(input) && player.shopUpgrades[input] < shopData[input].maxLevel) {
-                player.worlds.sub(getShopCosts(input));
-                player.shopUpgrades[input] += 1
-            }
+        p = await Confirm(`Are you sure you'd like to purchase ${upgradeCounter}*${input} for ${format(totalCost)} Quarks? Press 'OK' to finalize purchase.`);
```
KhafraDev, line 380:
```suggestion
        p = await Confirm(`Are you sure you'd like to purchase ${upgradeCounter}*${input} for ${format(totalCost)} Quarks?`);
```

KhafraDev (approved):

src/Shop.ts
KhafraDev: also maybe a `[number] x [number]` instead of using an asterisk?

nmussy: Done

Pseudonian:
Hello, this is currently presenting problems with the current shop code. If the code can be made compatible I can look at merging it again.

PR #110, opened 2021-07-24
Fix for the c15 ascend exploit and "buy into debt" for Plat upgrades

AreYouKittenMe:
Ascending in c15 could let a player play the challenge with less than the specified corruption if corruption/completes were high enough to avoid dumping them out of the challenge, for a highly inflated expo. This change ensures the c15 all-11 corruption array is kept intact. As a side effect it seems to even prevent the player from accidentally completing c15 with an ascend.


AreYouKittenMe:
Not reviewed yet? In that case I might as well add a fix for the "go into debt" Abyssals in Platonic upgrades. Begone!

AreYouKittenMe:
Conflicts with new Plat upgrade cost (which fixed that issue anywa), so nuking this PR, will submit a new one for the c15 fix

PR #115, opened 2021-07-25
Calculation overflow bug fix

httpsnet:
Cap the calculation overflow bug caused by Keynesian Statue

When the tax is reduced to 0, the coin income becomes 0, and the game does not work properly
This is currently reported by some players over 1e40 on Platonic Cubes
With this fix the next coin income 0 bug will not cause Platonic Cubes up to 1e65-1e70
I think this patch will give you enough synergy durability in 2.5.6

The patch has the following changes

Taxes are always capped at a minimum of 1e-300
Supports Decimal for some calculations
Adding a semicolon and renaming variables

### DIFF
--- a/src/Tax.ts
+++ b/src/Tax.ts
@@ -7,10 +7,9 @@ import { CalcECC } from './Challenges';
 import { achievementaward } from './Achievements';
 
 export const calculatetax = () => {
-    let c = 0;
-    let e = 1;
-    let f = 1;
-    let compareC = 0;
+    let exp = 1;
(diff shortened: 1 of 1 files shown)


Pseudonian:
Hello, I'm currently getting an error 
EventListeners.ts:373
document.getElementById('toggleautoresearchmode').addEventListener('click', () => toggleAutoResearchMode())
Appears that 'toggleautoresearchmode' is null. If this PR can be updated for that then I can accept it.

PR #118, opened 2021-07-26
Visible ascension count

Cephyric-gh:
Ascension count is now handled in its own function, and the value is shown on the corruption tab as long as ascending would provide more than 1 to the ascension counter


strentax:
Yay thanks Ceph

Pseudonian:
Thank you for the PR Cephyric- works on local test.

PR #119, opened 2021-07-27
Fix c13 for some saves

AreYouKittenMe:
Some saves have data in player.challengecompletions[0] which messes with their ability to complete c13. Now properly only counts c1 to 10.


AreYouKittenMe:
Review: Please check #120 before this, it includes the same fix among several.

PR #121, opened 2021-07-28
Apply max rune level

strentax:
- declare global variable representing max rune level limit of 40000
- consider this limit when calculating current max rune level
- use this max limit when displaying levels on Runes tab
- consider current level vs max level when displaying EXP to next rune level


strentax:
merge resolved

PR #125, opened 2021-08-02
Fix for c15 ascend bug/exploit

AreYouKittenMe:
This fixes the behavior of ascend in c15 by ensuring corruptions get set as needed. No more accidental completions, no more insane expos from abusing them. Ascend will now behave like in every other asc challenge - reset the current attempt. That's all.


Pseudonian:
seems fine on my end. Thanks!

PR #127, opened 2021-08-04
Minor Fixes to Hepts + Corruptions

ForgottenAngel:
- Fixed Loadout 4 always loading corruption levels as 0 when loading the game normally rather than through importing.
- Fixed Hepteracts not using up other Cube types when crafting.
- Changed version references from 2.5.6 to 2.5.7.
- Changed dependency reference of break_infinity.js to use a version number instead so `npm install` actually works.

### DIFF
--- a/src/Hepteracts.ts
+++ b/src/Hepteracts.ts
@@ -115,7 +115,7 @@ export class HepteractCraft {
         for (const item in this.OTHER_CONVERSIONS) {
             if (typeof player[item as keyof Player] === 'number')
                 (player[item as keyof Player] as number) -= amountToCraft * this.OTHER_CONVERSIONS[item as keyof Player];
-            else if (Object.prototype.isPrototypeOf.call(Cube, player[item as keyof Player]))
+            else if (player[item as keyof Player] instanceof Cube)
                 (player[item as keyof Player] as Cube).sub(amountToCraft * this.OTHER_CONVERSIONS[item as keyof Player]);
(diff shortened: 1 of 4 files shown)


src/Hepteracts.ts
```diff
@@ -115,7 +115,7 @@ export class HepteractCraft {
         for (const item in this.OTHER_CONVERSIONS) {
             if (typeof player[item as keyof Player] === 'number')
                 (player[item as keyof Player] as number) -= amountToCraft * this.OTHER_CONVERSIONS[item as keyof Player];
-            else if (Object.prototype.isPrototypeOf.call(Cube, player[item as keyof Player]))
+            else if (Object.prototype.isPrototypeOf(Cube))
```
KhafraDev, line 118: needs to be reverted

KhafraDev (changes requested):

src/Hepteracts.ts
ForgottenAngel: Out of curiosity, where does this cause issues? Unless having the hept crafts not take cubes was intentional, it seemed to work fine in testing.

ForgottenAngel: Correct that, I'm stupid and see the issue.

ForgottenAngel:
I realized that my previous change always evaluated as true because every object is prototype of Object. This new change should have actually fixed it.

1e1000000:
It should be major fix to Hept

PR #128, opened 2021-08-05
Allow 20 history records instead of 10.

GendoIkari82:


KhafraDev:
If the number of records is never changed, why is it on the `player` object in the first place?

GendoIkari82:
> If the number of records is never changed, why is it on the `player` object in the first place?

I was wondering that exact same thing when I saw the code! Thought that changing it would be going to far though. Maybe Kevin originally was planning on making it a setting or something?

KhafraDev:
@Pseudonian didn't add this feature in actually, and I don't believe he's touched the code here whatsoever since it was added. 

Anyways, I would prefer if this was added to the `Globals` object - https://github.com/Pseudo-Corp/SynergismOfficial/blob/af4de78ac5df1ac54aed39acd1df2110c84186c1/src/Variables.ts#L4 and deleted from the `player` object entirely.

GendoIkari82:
> @Pseudonian didn't add this feature in actually, and I don't believe he's touched the code here whatsoever since it was added.
> 
> Anyways, I would prefer if this was added to the `Globals` object -
> 
> https://github.com/Pseudo-Corp/SynergismOfficial/blob/af4de78ac5df1ac54aed39acd1df2110c84186c1/src/Variables.ts#L4
> 
> and deleted from the `player` object entirely.

Sounds good; done!

KhafraDev:
the key should be entirely deleted from any save being loaded. The best way to do this is to add `delete data.historyCountMax` in `checkVariablesOnLoad` function.

otherwise lgtm

GendoIkari82:
> the key should be entirely deleted from any save being loaded. The best way to do this is to add `delete data.historyCountMax;` in `checkVariablesOnLoad` function.
> 
> otherwise lgtm

I removed the property from the Player type completely; so it's already gone in the `checkVariablesOnLoad` function, where `data` is a Player type. I could remove it from the saved data inside loadSynergy, but it should just get removed as soon as the game next saves anyway. Can you please confirm that the a property being in the saved storage but not in the Player type won't cause any issues with the Mapping when loading?

kewne7768:
> If the number of records is never changed, why is it on the `player` object in the first place?

I figured some people would want a giant history and don't care about the savegames becoming huge, turned out that it's not that many people so I never built the actual setting UI for it.

Hardcoding a number should be fine at this point. I chose 10 as default because that's what would fit in the Kong iframe without a scrollbar, but that doesn't really matter anymore. 20 should be a good balance.

Pseudonian:
Looking good. Thank you for the PR!

PR #129, opened 2021-08-05
Fix Achievements 270

1e1000000:
It include:
Achievement 270 reward always getting applied
Achievement 270 reward scaling is 10x faster than intended

### DIFF
--- a/src/Calculate.ts
+++ b/src/Calculate.ts
@@ -1166,7 +1166,7 @@ export const calculateHepteractMultiplier = (score = -1) => {
         // Achievement 265 Bonus [Max: 160T Asc]
         1 + Math.min(0.2, player.ascensionCount / 8e14) * player.achievements[265],
         // Achievement 270 Bonus
-        Math.min(2, (1 + 1/100000 * Decimal.log(player.ascendShards.add(1), 10)))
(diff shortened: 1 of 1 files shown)


src/Calculate.ts
```diff
@@ -1166,7 +1166,7 @@ export const calculateHepteractMultiplier = (score = -1) => {
         // Achievement 265 Bonus [Max: 160T Asc]
         1 + Math.min(0.2, player.ascensionCount / 8e14) * player.achievements[265],
         // Achievement 270 Bonus
-        Math.min(2, (1 + 1/100000 * Decimal.log(player.ascendShards.add(1), 10)))
+        Math.min(2, (1 + 1/1000000 * Decimal.log(player.ascendShards.add(1), 10) * player.achievements[270]))
```
KhafraDev, line 1169:
```suggestion
        Math.min(2, (1 + 1/1_000_000 * Decimal.log(player.ascendShards.add(1), 10) * player.achievements[270]))
```

1e1000000: What was that, 1_000_000 seem weird and not working
Probably 1e6

1e1000000: But it was working :)

KhafraDev (approved):

src/Calculate.ts
KhafraDev: https://v8.dev/features/numeric-separators

PR #136, opened 2021-08-17
Remove “Most Hated Feature”

1e1000000:
They were a lot of users want to remove it, however, the confirmation of can’t refund item will remain always pop up :)


GendoIkari82 (approved):
Looks good to me; please merge this in!

PR #137, opened 2021-08-21
Fix 3 Achievements

1e1000000:
195 and 196: Fix achievement reward display capped at 2% instead of 25k% and 2k% respectively
270: Fix achievement reward display without getting capped at 100%


Pseudonian:
Looks good. Thank you for the PR.!

PR #143, opened 2021-09-01
Added Offering and Obtainium caps with ant rewards

httpsnet:
Add 1e300 to Offering and Obtainium caps with ant rewards
Currently Offering and Obtainium have a bug where the game freezes when it reaches 2 ^ 1024
Game destruction with 1e300 caps will be extended after at least 179769313 seconds

### DIFF
--- a/src/Calculate.ts
+++ b/src/Calculate.ts
@@ -745,8 +745,8 @@ export const calculateAntSacrificeRewards = (): IAntSacRewards => {
     const rewardsMult = G['timeMultiplier'] * G['upgradeMultiplier'];
     const rewards: IAntSacRewards = {
         antSacrificePoints: G['effectiveELO'] * rewardsMult / 85,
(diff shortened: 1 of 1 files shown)


1e1000000:
I think you can do after you this:
did a prestige reset offering to 1e300 if you have more than 1e300 of them
did a reincarnation reset obtainium to 1e300 if you have more than 1e300 of them
Also it can go past 1e300 by other reset, so you also need to hardcap the gain

httpsnet:
Yes. Exactly as you pointed out
I need to patch two additional files to fully hardcode the cap
But each of my pull requests can only send one file, which I thought would be annoying to developers

Offering and Obtainium never reach 2 ^ 1024 in the current version
The synergy resistance patch I do is to create a workspace after Singularity
It's up to the developer to ignore this

1e1000000:
What is the reason of closing?

PR #147, opened 2021-09-07
Some changes on Hept Forge and c15 desc

1e1000000:
Hide ‘If this is less than your input, you either hit the inventory limit or you had insufficient resources.’ If you craft max of them, because input didn’t even exist when crafting max.
Remove [WIP] on first Hept forge, it is implemented.


1e1000000:
Another comment is because c15 is still applied at Corruption Level 11 with a maximum of Level 13 available

PR #151, opened 2021-10-01
Add singularity to the cost color calculation for research

httpsnet:
Fixed singularityCount not being reflected in research calculations used for research cost style colors
Also added singularity to the calculation of research cost return when loading the game


httpsnet:
I confirmed the correction in the test version, so I will close it. Thank you

PR #160, opened 2022-04-09
2.9

KhafraDev:


Pseudonian:
Some people said it would never be done- others said it would take even longer. But this, April 10, 2022, is the end of the old era and the beginning of the end. Version 2.9, the second to last Synergism major release, is live! Thank you to all of the testers, all of the naysayers, and especially a big thanks to @KhafraDev who made this update possible! <3

PR #162, opened 2022-04-18
Update Platonic.ts

Xander374:


Xander374:
Update to p2x1 description.

PR #164, opened 2022-04-20
feat: use Blob to store saves

KhafraDev:
Using a blob should hopefully be more efficient and prevent some of the file from being cut off.

- Fixes a potential race condition where the game was closed after the save was deleted, but before the game was saved.
- Improve type safety by explicitly checking for `null`/`undefined` in some methods.

Main issues:
Safari v13 and older do not support `Blob.text()`


KhafraDev:
No safari 13- users reacted to https://discord.com/channels/677271830838640680/705896710630670336/966190835551047740 within ~24 hours so I'm gonna go ahead and merge this.

PR #174, opened 2022-05-17
Auto Challenge Sweep improvement

httpsnet:
If the challenge is the maximum number of completions, that challenge will be skipped.
If there are no more challenges to complete, exit Auto Challenge Sweep.
If the challenge reaches the maximum number of completions for the first time, the challenge will be exited.

### DIFF
--- a/src/Challenges.ts
+++ b/src/Challenges.ts
@@ -693,7 +693,7 @@ export const runChallengeSweep = (dt: number) => {
                     break;
                 }
             }
-            if (startChallenge == 11) {
+            if (startChallenge === 11) {
                 toggleAutoChallengeModeText('OFF');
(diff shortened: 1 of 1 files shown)


httpsnet:
For example, when you complete the Transcension Challenge up to 9001, you will automatically move to the next challenge.
Because of this lack of switching, I manually switched to the next challenge about every 10 seconds for each ascension.
It was stressful because it took a minute to complete this task. This patch can solve it.

KhafraDev (changes requested):

Pseudonian (approved):
I tested a number of different files and scenarios, all of which appear to work without error. Covered all changed scenarios. +1

PR #179, opened 2022-05-24
More Auto Challenge Sweep fix

httpsnet:
sorry. This time, we have fully verified the coding, so we will request it.

While running the Reincarnation Challenge, browse to player.highestchallengecompletions to not select the completed Transparent Challenge. This was a problem with my code I added.
If the next challenge is the same challenge, I will not restart that challenge.
Fixed to display the currently running mode correctly.
If all challenges are completed or not run, Auto Challenge Sweep will not stop on its own. This is because auto ascension may not work.


httpsnet:
This pull request solves all the bugs I know of.
It's painful to be pointed out in #bug-reports-please.

PR #188, opened 2022-06-07
bug fixes

httpsnet:
Changed to update the c15 score bonus on load before the calculation.
This will prevent the import from affecting the Obtainium bonus of the previous data.

Fixes that in cases like all levels 14 of Corruption, c10 can result in 0 coin production.
I could complete more than 100 challenges other than c10, but only c10 didn't move the coin with 2.
Even if the coin production ratio is 2, c10 can be easily achieved, so I think it is a problem.

### DIFF
--- a/src/Reset.ts
+++ b/src/Reset.ts
@@ -903,7 +903,7 @@ export const singularity = async (): Promise<void> => {
     hold.dailyCodeUsed = player.dailyCodeUsed
     hold.runeBlessingBuyAmount = player.runeBlessingBuyAmount
     hold.runeSpiritBuyAmount = player.runeSpiritBuyAmount
-    hold.transcendamount = player.prestigeamount
+    hold.prestigeamount= player.prestigeamount
     hold.transcendamount = player.transcendamount
     hold.reincarnationamount = player.reincarnationamount
     hold.talismanOne = player.talismanOne
--- a/src/Reset.ts
+++ b/src/Reset.ts
@@ -26,7 +26,7 @@ import { Synergism } from './Events';
 import type { Player, resetNames } from './types/Synergism';
 import { updateClassList } from './Utility';
 import { corrChallengeMinimum, corruptionStatsUpdate, maxCorruptionLevel } from './Corruptions';
-import { toggleAutoChallengeModeText } from './Toggles';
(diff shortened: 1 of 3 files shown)


KhafraDev (approved):

httpsnet:
Fixes an issue where free level calculations such as Cube Flame, Cube Blaze and Cube Inferno are not done.
Other files may also use the old method.
Search src for regular expression with "player.singularityUpgrades.(.+).level".

Tabs and subtabs are now initialized on loading.
The challenge summary is now displayed when loading.
Now writes a textbox on loading.
Now initializes settings that are not saved on loading.

More settings are inherited by singularity.

If anything happens, I will change it.

src/Reset.ts
```diff
@@ -870,13 +870,6 @@ export const singularity = async (): Promise<void> => {
     const hold = Object.assign({}, blankSave, {
         codes: Array.from(blankSave.codes)
     }) as Player;
-    //Reset Displays
-    toggleTabs('buildings');
-    toggleSubTab(1, 0);
-    toggleSubTab(4, 0); // Set 'runes' subtab back to 'runes' tab
-    toggleSubTab(8, 0); // Set 'cube tribues' subtab back to 'cubes' tab
-    toggleSubTab(9, 0); // set 'corruption main'
-    toggleSubTab(-1, 0); // set 'statistics main'
```
Pseudonian, line 879: If these don't reset you can end up having access to tabs which are otherwise not unlocked until later in the singularity. Keep these displays please

src/Reset.ts
```diff
@@ -908,6 +901,26 @@ export const singularity = async (): Promise<void> => {
     hold.autoResearchToggle = player.autoResearchToggle
     hold.autoResearchMode = player.autoResearchMode
     hold.dailyCodeUsed = player.dailyCodeUsed
+    hold.runeBlessingBuyAmount = player.runeBlessingBuyAmount
+    hold.runeSpiritBuyAmount = player.runeSpiritBuyAmount
+    hold.transcendamount = player.prestigeamount
```
Pseudonian, line 906:
```suggestion
    hold.prestigeamount= player.prestigeamount
```

src/Reset.ts
httpsnet: I returned it as you pointed out

PR #192, opened 2022-06-09
Fix c15 import/refresh bug

cloudytheconqueror:
There's a bug where if you import a save that's in c15 or refresh while inside c15, the ants become way quicker. When entering c15, the game sets all corruptions to level 11, including the removed corruption Divisiveness. This makes the sum of corruption levels 99, which raises ant production to the power of ^0.01. However if you import or refresh, the Divisiveness corruption is discarded, making the sum of corruption levels 88 and raising ant production to ^0.12, effectively raising ant production to the 12th power.

I thought that c15 applying a corruption that's no longer used is a bug in itself, so I made it no longer apply that corruption, and made it raise ant production to the power of 0.0833 (approximation of 1/12) to counteract the ant speedup mentioned above.

I also fixed the displayed current corruptions not being changed to 11 in the corruptions tab when entering c15, and also changed the displayed ant production in the corruptions tab to include nerfs applied by ascension challenges, and changed Maladaption to Viscosity in the ascend history.


Pseudonian:
This was differently fixed in a separate update. Thank you for the PR, though!

PR #199, opened 2022-06-16
DIsplay amount of GQ gained in ResetInfo text

jarboo:
Display the number of Golden Quarks gained on Singularity in the flavor text. It uses the same function that displays it in the Dialogs when confirming Singularity reset

### DIFF
--- a/Synergism.css
+++ b/Synergism.css
@@ -697,15 +697,14 @@ nav li:hover {
     margin-right: 5px;
 }
 
-#promocodes,
-#theme {
+.promocodeButton {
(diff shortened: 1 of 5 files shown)


jarboo:
image

jarboo:
Originally I wanted to create 2 separate Pull Requests, but it didn't work as I thought.
So this PR contains 2 features : 
- display GQ as explained above
- add 3 new buttons in the Settings tab to have direct access to "daily", "add", and "time" codes. It looks like this
image

PR #205, opened 2022-06-17
Added a toggle to disable Singularity confirmation

jarboo:
The new toggle appears when both singulartiyCount and ascensionCount are > 0 because when entering a new Sing, confirmations for Ascension and Ant Sac are hidden.
When the confirmation is OFF, clicking the Singularity button triggers only 1 Confirmation instead of 5 Alerts and 3 Confirmations when it's ON
image
image

### DIFF
--- a/src/CheckVariables.ts
+++ b/src/CheckVariables.ts
@@ -214,7 +214,7 @@ export const checkVariablesOnLoad = (data: PlayerSave) => {
         player.toggles[32] = true;
     }
 
-    if (data.toggles[33] === undefined) {
+    if (!data.toggles || data.toggles[33] === undefined) {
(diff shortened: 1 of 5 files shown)


src/CheckVariables.ts
```diff
@@ -214,6 +214,10 @@ export const checkVariablesOnLoad = (data: PlayerSave) => {
         player.toggles[32] = true;
     }
 
+    if (data.singularityCount === 0) {
+        player.toggles[33] = true;
+    }
```
Pseudonian, line 219:
```suggestion
    if (data.toggles[33] === undefined) {
        player.toggles[33] = true;
    }
```

PR #206, opened 2022-06-18
Minor fix

httpsnet:
Fixed a display issue with #settingstab.
I updated the HTML of each resource only when the textContent changed. This should significantly reduce the page refresh process.
Fixed an issue with achievement messages.
Added "date" to promotion code.
The promotion code allows uppercase letters. The button text in Special Actions should work as is.
spoiler added octeracts per second.

### DIFF
--- a/src/ImportExport.ts
+++ b/src/ImportExport.ts
@@ -426,7 +426,7 @@ export const promocodes = async (input: string | null) => {
     } else if (input === 'time') {
         const availableUses = timeCodeAvailableUses();
         if (availableUses === 0) {
-            return Alert(`
+            return Confirm(`
             If you imported a save, you cannot use this code for 15 minutes to prevent cheaters.
             
             Regardless, you must wait at least 15 minutes between each use.
--- a/src/ImportExport.ts
+++ b/src/ImportExport.ts
@@ -190,14 +190,16 @@ export const promocodesInfo = async (input: string) => {
             availableUses = addCodeAvailableUses();
             textMessage += availableUses + ' use' + (availableUses !== 1 ? 's' : '') + ' left.';
             if (availableUses === 0) {
-                textMessage += ' Next: in ' + format(addCodeTimeToNextUse(), 2, true) + ' seconds.';
+                textMessage += ' Next: in ' + format(addCodeTimeToNextUse(), 0, true) + ' seconds.';
             }
             break;
         case 'time':
             availableUses = timeCodeAvailableUses();
-            textMessage += availableUses + ' use' + (availableUses !== 1 ? 's' : '') + ' left. reward x' + format(timeCodeRewardMult(), 2, true) + '.';
+            textMessage += availableUses + ' use' + (availableUses !== 1 ? 's' : '') + ' left.';
             if (availableUses === 0) {
-                textMessage += ' Next: in ' + format(timeCodeTimeToNextUse(), 2, true) + ' seconds.';
+                textMessage += ' Next: in ' + format(timeCodeTimeToNextUse(), 0, true) + ' seconds.';
+            } else {
+                textMessage += ' Reward x' + format(timeCodeRewardMult(), 2, true);
             }
             break;
         default:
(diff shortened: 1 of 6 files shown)


src/ImportExport.ts
```diff
@@ -475,7 +478,14 @@ export const promocodes = async (input: string | null) => {
 
         const ascensionSpeed = calculateAscensionAcceleration()
         const perSecond = 1/(24 * 3600 * 365) * baseMultiplier * productContents(valueMultipliers) * ascensionSpeed
-        return Alert(`You will gain an octeract (when they come out) every ${format(1 / perSecond, 2, true)} seconds, assuming you have them unlocked!`)
+        const perSecondText = perSecond >= 1 ? `${format(perSecond, 2, true)} per second` : `every ${format(1 / perSecond, 2, true)} seconds`
+        return Alert(`You will gain an octeract (when they come out) ${perSecondText}, assuming you have them unlocked!`)
+    } else if (input === 'date') {
```
KhafraDev, line 483: is this needed?

KhafraDev (changes requested):

src/ImportExport.ts
httpsnet: Before Cx20 was nerfed, I've been able to get less than 0.1 with normal data.
Therefore, I think it may progress to 0 in future versions.
Ascension Speed is too powerful and I think it should be reduced by exponentiation.
If it's still not needed, I'll erase the code I pointed out.

httpsnet: I got 0.72 seconds with spoiler at 10/14/14/14/14/14/14/14 of main data sing 33 in v2.9.8.
Upload data if necessary.
Should the date code be erased? I think Synergism needs a watch.
The following ImportExport.ts upload contains more fixes.

KhafraDev: I don't think adding this as a promocode is the solution

httpsnet: ok, deleted

KhafraDev (changes requested):

src/ImportExport.ts
```diff
@@ -421,22 +424,25 @@ export const promocodes = async (input: string | null) => {
     } else if (input === 'time') {
         const availableUses = timeCodeAvailableUses();
         if (availableUses === 0) {
-            return Confirm(`
+            return Alert(`
```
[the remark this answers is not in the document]
httpsnet: All the questions you point out are confusing to users, as the results are the same for yes and no.
When I actually confirmed it by the user's report, I understood it.

src/ImportExport.ts
```diff
@@ -424,22 +424,25 @@ export const promocodes = async (input: string | null) => {
...
             If you imported a save, you cannot use this code for 15 minutes to prevent cheaters.
```
httpsnet, line 428: Here, fraudster measures aren't really working.
Should I take measures like gamble's localStorage?

src/ImportExport.ts
```diff
@@ -427,16 +432,18 @@ export const promocodes = async (input: string | null) => {
...
 
         const random = Math.random() * 15000; // random time within 15 seconds
         const start = Date.now();
-        await Confirm(
-            'Click the button within the next 15 seconds to test your luck!' +
+        const confirmed = await Confirm(
+            'Click the OK button within the next 15 seconds to test your luck!' +
             ` If you click within ${format(2500 + 125 * player.cubeUpgrades[61], 0, true)} ms of a randomly generated time, you will win a prize!` +
-            ` This particular instance has a ${format(rewardMult, 2, true)}x multiplier due to elapsed time between uses.`
+            ` This particular instance has a x${format(rewardMult, 2, true)} multiplier due to elapsed time between uses.`
```
KhafraDev, line 442: x should be at the back
"5.5x"
"x5.5"

KhafraDev (changes requested):

src/ImportExport.ts
httpsnet: ok

src/ImportExport.ts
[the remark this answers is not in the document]
httpsnet: I didn't get a reply, so I accepted your request and resolved it

httpsnet:
I fixed more. sorry.

src/Toggles.ts
```diff
@@ -47,6 +47,9 @@ export const toggleTabs = (name: keyof typeof tabNumberConst) => {
     revealStuff();
     hideStuff();
 
+    const el = document.activeElement as HTMLElement;
```
KhafraDev, line 50: this is nullable

src/Toggles.ts
```diff
@@ -321,16 +324,22 @@ export const keyboardTabChange = (dir = 1, main = true) => {
 }
 
 export const toggleSubTab = (mainTab = 1, subTab = 0) => {
-    if (tabs(mainTab).unlocked && subTabsInMainTab(mainTab).subTabList.length > 0) {
+    const subTabs = subTabsInMainTab(mainTab)
+    if (tabs(mainTab).unlocked && subTabs.subTabList.length > 0) {
+
+        const el = document.activeElement as HTMLElement;
```
KhafraDev, line 330: ditto

KhafraDev (changes requested):

src/Toggles.ts
httpsnet: ok
"eslint-disable-next-line @typescript-eslint/no-unnecessary-condition" is also gone.

httpsnet:
I forgot to write the update contents of ebb13ed

Fixed an issue where all cube upgrades would be initialized when reloading after singularity and before ascension
This was the cause of the loss of w1x7 earned in updateSingularityMilestoneAwards().
Added w1x7 with singularity 20. Added w1x10 in singularity 11.
Auto Ascend fixed an issue that could actually be done just by enabling the toggle.

PR #212, opened 2022-06-20
Special Action Time improvements

jarboo:
Fixed cancel button not working
Added Multiplier to info text
Removed Promocode button hover event because it sometimes cleared the message after inputing a real promocode, before being able to read it


httpsnet:
It's similar to what I changed in #206. Either one will be a conflict...

jarboo:
> It's similar to what I changed in #206. Either one will be a conflict...

Oh yeah, you're right :( 
I didn't see your changes.
Be careful: your PR is called Minor Fix and contains a LOT of different things.
When possible, you should create a different branch and PR for each subject. If your branch had been called "Special Action time something", I would have seen it ;)

KhafraDev (approved):

PR #218, opened 2022-06-21
Add 4 new corruption Loadouts

krynczak:
This is a work-in-progress.

It works just as expected, the only thing not done yet is finding out why the new loadouts are named as `undefined` (you can rename then and it work just as the other ones do).


krynczak:
Closed because KhafraDev just did a new PR.

PR #221, opened 2022-06-21
Singularity count in save file

awWhy:
Dont ask how long it took
Everything might explode, cause I have no idea


awWhy:
This how it looks if you save with it
"Synergism-v2.9.8 Decanter-2022-06-21 10_57_01-Singularity 39"

PR #224, opened 2022-06-21
Support dynamic corruption loadouts

Cowboy-coder:
New features:
- Add new Corruption loadout row
- Remove corruption loadout row
- Corruption loadout is scrollable. (kinda ugly but maybe it's fine.)

Screenshot 2022-06-21 at 20 09 35


src/Corruptions.ts
```diff
@@ -204,69 +202,114 @@ export const corruptionButtonsAdd = () => {
...
-            cell.title = 'Reset corruptions to zero on your next ascension'
-        } else {
-            let cell = row.insertCell();
-            let btn = document.createElement('button');
-            btn.className = 'corrSave'
-            btn.textContent = 'Save'
-            btn.onclick = () => corruptionLoadoutSaveLoad(true, i);
-            cell.appendChild(btn);
+    return `
+      <table id="corruptionLoadoutTable" alt="corruptionLoadoutTable" label="corruptionLoadoutTable">
```
Cowboy-coder, line 213: Not so optimal to just replace-it-all-in-place on each "update". But I think this should be fine for this view. Manually keeping track of updating the DOM seems more annoying. (this is where I miss a small lightweight react-like library :P)

KhafraDev (changes requested):
needs to be discussed

httpsnet:
Thank you for your wonderful ideas and efforts. But discussion is needed.

Please confirm that the Remove operation does not cause any mistakes.
I don't think it's a good implementation to include the HTML in loadoutTemplate() directly in your code.
The hotkey Shift + Number is a loadout change. It corresponds from 0 to 9.
There is a conflict in # 219. Be careful.

Cowboy-coder:
@httpsnet I don't like the fact that the html is embedded in the js either. But manually keeping track of the DOM for new/save/load/remove/rename is kinda tricky and error prone.

PR #230, opened 2022-06-22
Uppercase nouns

awWhy:
Will keep this PR as only descriptions changes, mostly with turning all small letters into big where it needs to be. But with some other text fixes.


awWhy:
Goint to move Disable alert for buying p4x5 into another PR

PR #233, opened 2022-06-23
Improve singularity tab and golden quarks display

jarboo:
A more detailed Singularity Tab with 3 subtabs:
- Shop : same as the current Upgrades, with a nice icon for golden quarks
- Penalties : same as the current Penalties listed above the upgrades. In the future, we could display the current penalties and the penalties of next Sing side by side. And maybe add a warning when the next Sing is especially difficult
- Perks : new feature. Equivalent to the current last column of achievements + the different singularity milestones that are not displayed anywhere in the game.
image
image
image

Nothing is definitive (names, GUI, etc.)
I've also added the singularity count in the header, so that it's always visible at a glance and can be shared with others with the sreenshots


jarboo:
I wanted to modify the Penalties subtab a bit more, but since Plat is about to change Singularity balance, with new penalties introduced  down the line, I think it's better to merge this sooner than later, if it's ok with you :)

PR #240, opened 2022-06-25
History visuals and New Toggle

awWhy:
Not going to bother with changin Big Letters as Khafra wont like it


awWhy:
If this PR will be merged, will add another PR for new toggle to disable Hepteracts alerts (Not expand one)
Cant do this before it as will create conflicts, and better not here, as might be buggy

KhafraDev (approved):

PR #260, opened 2022-06-29
Themes Rework

awWhy:
1 more theme and Light mode rework, here colors for comparison
image (Newest) image

Mikkyne suggested Ach text color, Light and Light Dark mode colors

### DIFF
--- a/Synergism.css
+++ b/Synergism.css
@@ -31,6 +31,10 @@ body.dark {
     background-color: #111;
 }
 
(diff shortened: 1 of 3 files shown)


awWhy:
Added some more colors
And temparary changes to dark mode (Only if go to it through theme swithing), if people will like it can be done Permanent
image

awWhy:
Force push because forgot to change Button name...

PR #261, opened 2022-06-30
Prevents the daily refill timer from bugging out to a future date

marlyn-x86:
`performance.now()` does not always return to 0 on a reload, but `loadingDate` does. By instead taking the difference between `performance.now()` measurements for getting time pinned to the loading date, we can prevent the time from getting pushed into the future and causing missed refill timers.

This was also what was causing people to report refill timer values converging on times other than their local midnight -  this misaligned time is a symptom of the perceived date being pushed into the future, with their `dayCheck` value coming along with it.


KhafraDev (approved):
> performance.now() does not always return to 0 on a reload

Gotta love chrome

Pseudonian:
Brilliant work discovering the cause of the bug!

PR #262, opened 2022-06-30
Instant Challenge 2 and new Toggle to hide maxed upgrades

awWhy:
My branch for this is a bit old (But doesnt look like any conflicts)
It will nerf early Singularities ticks per Second (Like 1-5 less per tick),
But will buff ticks for c13 (only R challenges)
Leaving Ascension challenge now does cause Reset (Nillo asked)
Unlocked with shop Bonanza, cost 2e6 Quarks (Placeholder)
And finnaly new border colors for Liquid shop and also I moved them (At least away from buying GQ button)


awWhy:
Dont merge yet, will try to add || manual (Nillo asked) so leaving challenge would actually cause reset

KhafraDev (approved):

awWhy:
Sorry for putting stuff in legacy. I was thinkin that upgrade.bough is another constant, so I added without noticing it was legacy file

PR #265, opened 2022-06-30
Dracula theme

awWhy:
Most unique theme. Cant finish it though, because of how code for most of buttons being outside of css...
Should be anyway in good condition to be merged.


awWhy:
Fix for wrong border for shop in Ligher Dark Mode

awWhy:
Didnt deleted, just updated branch name...

PR #267, opened 2022-07-01
Fix being able to open fractions of Cubes

awWhy:
It will give in average 0,6% less cubes now, but no getting free cubes (Like 10 per second)
It fixes both my buttons and Khafra percentage feature


awWhy:
Didnt deleted this one either, only changed branch name, because I was asked...

PR #268, opened 2022-07-01
Added missing perks and levels

jarboo:
Added perk data for existing effects:
- Golden coins: GC boost coin
- Generous Orbs: Orb effect for opening cubes
Extra level for Super Start

### DIFF
--- a/src/singularity.ts
+++ b/src/singularity.ts
@@ -20,7 +20,7 @@ export const updateSingularityPenalties = (): void => {
 }
 
 function getSingularityOridnalText(singularityCount: number): string {
-    return 'You are in the <span style="color: gold">' + toOrdinal(singularityCount) + ' singularity</span>';
+    return 'You are in the <span style="color: gold">' + toOrdinal(singularityCount) + ' Singularity</span>';
 }
 
 export interface ISingularityData {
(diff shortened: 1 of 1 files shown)


src/singularity.ts
```diff
@@ -679,14 +679,48 @@ export const singularityPerks: SingularityPerk[] = [
...
+        levels: [1],
+        description: () => {
+            return 'Unspent Golden Quarks boost Coin gain. Especially strong for first Ascensions of each Singularity.'
+        }
+    },
+    {
+        name: 'Generous Orbs',
+        levels: [1, 2, 5, 10, 15, 20, 25, 30, 35],
+        description: (n: number, levels: number[]) => {
+            if (n >= levels[8]) {
```
KhafraDev, line 693: this entire section could be done in a loop if you change the bonus percent to be in an object

```js
const overfluxBonus = {
  8: 700,
  // ...
} as const;

for (let i = 8; i > 0; i--) {
  if (n >= levels[i]) {
    return `Overflux Orbs effect on opening Cubes for Quarks can now go up to ${overfluxBonus[i]}`
  }
  
  return 'Overflux Orbs effect on opening Cubes for Quarks can now go up to 215%'
}

jarboo: I agree and I hesitated to implement it like this. It might look "sexier", but is it really more performant or easier to understand for the next dev who will have to change this function? Even the total number of characters in the code file will not be that different. I don't know, it's your call :)

KhafraDev: still readable and easier to add onto (& less duplicate code output in the bundle)

KhafraDev (approved):

PR #271, opened 2022-07-02
More QoL

awWhy:
New toggle to auto buy Fragments (s40 and Cx1 required)
New toggle to save offerings (Just turns off auto runes)
Moved AoAG in to this PR (s50)
New hotkeys to use potions


awWhy:
No idea why github need me to hold its hand to solve this type of conflict
image

PR #273, opened 2022-07-02
Fix being able to open fractions of Cubes

awWhy:
You can open Cubes with 10%, 50% and custom amount (if %), that ends up being round up
Giving you free cubes openings (I was able to get 4k tess opened from only having 6)
This should fix it, but will decrease by ~0,6% amount of cubes you get from using these buttons

Also I didnt deleted previous PR, I only changed name of branch...

### DIFF
--- a/src/Ants.ts
+++ b/src/Ants.ts
@@ -340,7 +340,7 @@ export const sacrificeAnts = async (auto = false) => {
     let p = true
 
     if (player.antPoints.gte('1e40')) {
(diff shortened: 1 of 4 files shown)


awWhy:
Really old bug looks like, you can enter c15 or higher corruption level and still retain diamonds you have
Only if you have p3x1 (Will have same amount of Diamonds as Particles, before you Ascended)

KhafraDev (approved):

PR #274, opened 2022-07-03
made it so divine pack has multiplier of 7.77x instead of 2x in spoiler code and stat page

ab3327:
Divine pack multiplier was being incorrectly calculated as 2x multiplier.


ab3327:
In your updatedOcteract branch you are still using the incorrect multiplier in octeractGainPerSecond. This means you are likely balancing around the incorrect multiplier.

KhafraDev (approved):

PR #281, opened 2022-07-10
Auto hept ratios

ab3327:


ab3327:
just found a few bugs so definitely don't merge yet

awWhy:
Probably better formula would be like this:
Hepteracts to spend / all ratios added to each other * this hepteract ratio
Ratio minimum have to be 1 (Obviously why)

awWhy:
This way you also dont need to care which hepteract has biggest ratio

awWhy:
Hover priorities on all of these buttons are broken
image

awWhy:
Double Singularity bug wasnt stopped, was only made it easier to do
As it doesnt reset rune 7 anymore

PR #282, opened 2022-07-12
Heteract craft keep auto settings on sing

jarboo:
Auto settings are kept on Singularity.
Auto can be set ON/OFF even when craft is not unlocked but it triggers only when unlocked.
Display that a craft is locked in the colored progress bar.
Removed autoHepteractCount global variable because it created unnecessary complexity: it was updated in several parts of the game but used only once -> replaced by function getAutoHepteractCrafts()
image

image

### DIFF
--- a/src/Hepteracts.ts
+++ b/src/Hepteracts.ts
@@ -257,24 +257,16 @@ export class HepteractCraft {
         return this;
     }
 
-    toggleAutomatic(bool?: boolean): Promise<void> | HepteractCraft {
+    toggleAutomatic(newValue?: boolean): Promise<void> | HepteractCraft {
+        const HTML = DOMCacheGetOrSet(`${this.HTML_STRING}HepteractAuto`);
 
-        if (!this.UNLOCKED && bool === undefined) {
-            return Alert('You do not have this as an unlocked craft. Automation is therefore not possible.')
-        }
-        const HTML = DOMCacheGetOrSet(`${this.HTML_STRING}HepteractAuto`)
+        // When newValue is empty, current value is toggled
+        this.AUTO = newValue ?? !this.AUTO;
 
-        this.AUTO = bool ?? !this.AUTO
+        HTML.textContent = `Auto ${this.AUTO ? 'ON' : 'OFF'}`;
+        HTML.style.border = `2px solid ${this.AUTO ? 'green' : 'red'}`;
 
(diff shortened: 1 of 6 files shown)


src/Hepteracts.ts
```diff
@@ -572,6 +563,21 @@ export const overfluxPowderWarp = async () => {
     }
 }
 
+/**
+ * Get the HepteractCrafts which are unlocked and auto = ON
+ * @returns Array of HepteractCraft
+ */
+export const getAutoHepteractCrafts = () => {
+    const autoHepteracts: HepteractCraft[] = [];
+    for (const craft in player.hepteractCrafts) {
```
KhafraDev, line 572: use a for of with Object.keys instead

src/Reset.ts
```diff
@@ -956,6 +954,15 @@ export const singularity = async (): Promise<void> => {
     hold.researchBuyMaxToggle = player.researchBuyMaxToggle
     hold.cubeUpgradesBuyMaxToggle = player.cubeUpgradesBuyMaxToggle
 
+    // Quark Hepteract craft is saved entirely. For other crafts we only save their auto setting
+    hold.hepteractCrafts.quark = player.hepteractCrafts.quark;
+    for (const craftName in player.hepteractCrafts) {
```
KhafraDev, line 959: ditto

src/Reset.ts
```diff
@@ -956,6 +954,15 @@ export const singularity = async (): Promise<void> => {
...
+        if (craftName !== 'quark' ) {
```
KhafraDev, line 960:
```suggestion
        if (craftName !== 'quark') {
```

src/UpdateVisuals.ts
```diff
@@ -430,26 +430,34 @@ export const visualUpdateCubes = () => {
...
-
-    let barColor = '';
-    if (barWidth < 34) {
-        barColor = 'red';
-    } else if (barWidth >= 34 && barWidth < 68) {
-        barColor = '#cca300';
+    const text = type + 'ProgressBarText';
+    const bar = type + 'ProgressBar';
+    const textEl = document.getElementById(text)!;
+    const barEl = document.getElementById(bar)!;
```
KhafraDev, line 436:
```suggestion
    const textEl = document.getElementById(text)!;
    const barEl = document.getElementById(bar)!;
```
use DOMCacheGetOrSet

KhafraDev (changes requested):

PR #283, opened 2022-07-12
fixed accidental removal of /s from asc stats at top

ab3327:
oops


src/Toggles.ts
```diff
@@ -927,7 +927,9 @@ export const toggleAscStatPerSecond = (id: number) => {
     }
 
     el.textContent = player.ascStatToggles[id] ? '/s' : '';
-    el.textContent = player.ascStatToggles[6] ? '' : '';
+    if (id === 6){
+        el.textContent = player.ascStatToggles[id] ? '' : '';
```
KhafraDev, line 931:
```suggestion
    if (id === 6) {
        el.textContent = '';
```

are you sure this is correct?

ab3327: both toggles

PR #284, opened 2022-07-13
New mode for Auto Ascension

awWhy:
New mode for Auto Ascension - Real-time (Not affected by adds)
To make this one more usefull (Can still be used with MD14, with "Time before entering challenge"), need to add better w3x8 (I liked idea of rewarding it with aschievement)
Auto Ascension no longer possible if inside Asc Challenge (Since it will only hurt to Auto Ascend inside)

(Pushed to fix, because Github auto removed const that had nothing to do with this PR)

### DIFF
--- a/src/Helper.ts
+++ b/src/Helper.ts
@@ -31,12 +31,13 @@ export const addTimers = (input: TimerInput, time = 0) => {
             player.reincarnationcounter += time * timeMultiplier;
             break;
         }
-        case 'ascension': {
+        case 'ascension': { //Anything in here is affected by add code
             player.ascensionCounter += time * timeMultiplier * calculateAscensionAcceleration();
             player.ascensionCounterReal += time * timeMultiplier
             break;
         }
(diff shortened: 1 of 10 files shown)


src/Helper.ts
```diff
@@ -31,12 +31,13 @@ export const addTimers = (input: TimerInput, time = 0) => {
             player.reincarnationcounter += time * timeMultiplier;
             break;
         }
-        case 'ascension': {
+        case 'ascension': { //Anything in here is affected by add code (Dumb)
```
KhafraDev, line 34:
```suggestion
        case 'ascension': { //Anything in here is affected by add code
```

KhafraDev (changes requested):

src/Helper.ts
awWhy: (Still think its dumb)

KhafraDev (approved):

PR #245, opened 2022-06-26
Makes cube based researches permanent after sing0

LosNillos:

PR #285, opened 2022-07-13
Auto Opening for Cubes

awWhy:
Auto openings for cubes
For it to be more usefull need some permanent researches for tributes, tess and hypers (Like r8x8)
If r8x8 will become permanent, then better to add for auto Hypers check if it was bought (To prevent Hypers being wasted)


Synergism.css
```diff
@@ -2143,6 +2143,30 @@ nav li:hover {
     background-color: #333;
 }
 
+.autoOpensBtn {
+    position: absolute;
```
KhafraDev, line 2147: doesn't need to be absolute positioned

src/Reset.ts
```diff
@@ -649,6 +666,22 @@ export const reset = (input: resetNames, fast = false, from = 'unknown') => {
                 }
             }
         }
+
+        //Auto open Cubes. If to remove !=0, game will lag a bit if it was set to 0
+        if (player.autoopenCubes && player.openCubes != 0 && player.cubeUpgrades[51] > 0) {
```
KhafraDev, line 671: !==

src/Reset.ts
```diff
@@ -649,6 +666,22 @@ export const reset = (input: resetNames, fast = false, from = 'unknown') => {
...
+            player.wowCubes.open(Number(player.wowCubes) * player.openCubes / 100, false)
+        }
+        if (player.autoopenTesseracts && player.openTesseracts != 0) {
```
KhafraDev, line 674: !==

src/Reset.ts
```diff
@@ -649,6 +666,22 @@ export const reset = (input: resetNames, fast = false, from = 'unknown') => {
...
+            if (player.tesseractAutoBuyerToggle != 1 || player.resettoggle4 === 2) {
```
KhafraDev, line 675: !==

src/Reset.ts
```diff
@@ -649,6 +666,22 @@ export const reset = (input: resetNames, fast = false, from = 'unknown') => {
...
+                player.wowTesseracts.open(Number(player.wowTesseracts) * player.openTesseracts / 100, false)
+            }
+        }
+        if (player.autoopenHypercubes && player.openHypercubes != 0) {
```
KhafraDev, line 679: !==

src/Reset.ts
```diff
@@ -649,6 +666,22 @@ export const reset = (input: resetNames, fast = false, from = 'unknown') => {
...
+            player.wowHypercubes.open(Number(player.wowHypercubes) * player.openHypercubes / 100, false)
+        }
+        if (player.autoopenPlatonicsCubes && player.openPlatonicsCubes != 0) {
```
KhafraDev, line 682: !==

Synergism.css
awWhy: It wont move with zoom, dont worry

awWhy:
If this ever going to be merged, it will need small changes to allow to work with themes
Will do it a bit later, after done with themes rework

Synergism.css
awWhy: Its no longer is absolute positioned, but no idea if my way is any better than absolute,
At least it still doesnt move text

awWhy:
#245 
I for most effect I need that in any way to be in game, also to add check if r8x8 is bough for auto opens of Hypers
(I cant right now because then Hypers will never we be opened)

src/types/Synergism.d.ts
```diff
@@ -408,6 +408,14 @@ export interface Player {
     ascensionCounter: number
     ascensionCounterReal: number
     ascensionCounterRealReal: number
+    autoopenCubes: boolean,
```
KhafraDev, line 411: camelcase this: 'autoOpenCubes'

src/types/Synergism.d.ts
```diff
@@ -408,6 +408,14 @@ export interface Player {
...
+    openCubes: number
+    autoopenTesseracts: boolean,
```
KhafraDev, line 413: same

src/types/Synergism.d.ts
```diff
@@ -408,6 +408,14 @@ export interface Player {
...
+    openTesseracts: number
+    autoopenHypercubes: boolean,
```
KhafraDev, line 415: same

src/types/Synergism.d.ts
```diff
@@ -408,6 +408,14 @@ export interface Player {
...
+    openHypercubes: number
+    autoopenPlatonicsCubes: boolean,
```
KhafraDev, line 417: same

KhafraDev (changes requested):

KhafraDev (approved):

src/Reset.ts
```diff
@@ -650,6 +667,22 @@ export const reset = (input: resetNames, fast = false, from = 'unknown') => {
...
+        }
+        if (player.autoOpenTesseracts && player.openTesseracts !== 0) {
+            if (player.tesseractAutoBuyerToggle !== 1 || player.resettoggle4 === 2) {
+                player.wowTesseracts.open(Number(player.wowTesseracts) * player.openTesseracts / 100, false)
+            }
+        }
+        if (player.autoOpenHypercubes && player.openHypercubes !== 0) {
+            player.wowHypercubes.open(Number(player.wowHypercubes) * player.openHypercubes / 100, false)
+        }
+        if (player.autoOpenPlatonicsCubes && player.openPlatonicsCubes !== 0) {
```
mustache0110, line 683: You're testing for player.cubeUpgrades[51] because it unlocks automation in Cube tab?

But you're only testing it on WoW! Cubes and not Tessseracts/Hyper/Platonic. Do you intend to move the automation test up? Or apply it to each cube in turn?

Also, you're potentially opening partial cubes unless you floor the value you send to .open( )

And is it worth waiting for researches to open cubes? You could move autobuy cubes out of Reset and into AutoBuyCubes() function and have the buyResearch function call it when R8x23 (198) is maxed. Something like:
```
if (isResearchMaxed(index)) {
    DOMCacheGetOrSet(`res${player.autoResearch || 1}`).classList.remove('researchRoomba');
    if (index === 198 && player.cubeUpgrades[51] > 0 ) {
        AutoBuyCubes();
    }
}
```

awWhy: Well, its ment to happen only after Ascension.
Cube upgrade 51 (Cx1), was added only to 3d cubes, to prevent accidental oppening of all cubes
(For first Ascension, as that might hurt player)
Makin value floor might make calculation slower (?), exploiting to Ascend to open 1 extra cube will only hurt
And for researches, might easier solution is to make research like r8x23 permanent and not reset on Ascension
(But because of bugs researchs still have effects shortly after being erased, so you wont lose anything)

PR #286, opened 2022-07-13
Smart Loadout (Auto corruptions)

awWhy:
Smart Loadout - Sorts of auto Corruptions (Stops workin after p4x5, because reasons)
Has a tiny bug - it will still apply corruptions (for 1 Ascension, Asc Ch one) right after leaving ascension Challenge
I tryed to make all 3 tabs to look same, but it made corruptions to be at top instead (Might look weird, when first unlock)
Small bug fix for allowing to bring lvl 14 corr into save that never had corruptions saved


src/Synergism.ts
```diff
@@ -631,7 +631,28 @@ export const player: Player = {
...
+    smartLoadoutNames: [
+        'c14', //No idea how to call them
+        'r8x25', //It wont self update names, if to change them
+        'w5x10',
+        'p2x1',
+        'p3x2',
+        'Omega',
+        'p4x2',
+        'Challenges'
+    ],
```
KhafraDev, line 653: Remove the loadout names and instead have it be the object keys of smartLoadouts

KhafraDev (changes requested):

src/Synergism.ts
awWhy: I dont fully understand, I tryed this (I removed player.smartLoadoutNames from everywhere)
image

PR #291, opened 2022-07-14
Smart Loadout (Reopen)

awWhy:
Plat accidently merged

Smart Loadout - Sorts of auto Corruptions (Stops workin after p4x5, because reasons)
Has a tiny bug - it will still apply corruptions (for 1 Ascension, Asc Ch one) right after leaving ascension Challenge
I tryed to make all 3 tabs to look same, but it made corruptions to be at top instead (Might look weird, when first unlock)
Small bug fix for allowing to bring lvl 14 corr into save that never had corruptions saved


src/Synergism.ts
```diff
@@ -632,7 +632,18 @@ export const player: Player = {
         'Loadout 7',
         'Loadout 8'
     ],
-    corruptionShowStats: true,
+    smartLoadouts: {
```
KhafraDev, line 635: What I meant here was to change the number keys to the names of the loadout:

```js
smartLoadouts: {
  name: [...],
  otherName: [...]
}
```

KhafraDev (changes requested):

src/Synergism.ts
awWhy: That will be really complicated, as its using [number, number()], changin first to string will cause a lot of conflicts
Feels like a lot of work since names arent meant to be changed

awWhy: And I would still have to include line for building names, so it wont save any space

awWhy: image
image
image
With my knowledge that would require a lot of changes, for a no real benefit

KhafraDev: It's extremely easy because `smartLoadouts` only exists on this branch so no migrations need to be done. The issue is that your local branch has numbered keys; you need to clear your current save and reload one that hasn't been imported yet.

However if the names are useless, smartLoadouts should be an array.

awWhy: Clearing cookies doesnt help
Will try array

awWhy:
Will change Reqs to s25 and add some Sugestions

src/Synergism.ts
awWhy: Numbers are being used now, also removing !important from save button, broke .hover if you ever switch theme in current session (Same with delete button)

KhafraDev: You want an array of number[][], aka "an array of an array of numbers"
```js
smartLoadouts: [
    [0, 1, 2, 3, 4],
    [5, 6, 7, 8, 9]
]
```

awWhy: I tryed both of these methods, I wasnt able to past through past offline screen
And even if I do make it work, I still would have almost same code line, since game doesnt auto creates names

awWhy:
If this ever going to be merged, it will need small changes to allow to work with themes
Will do it a bit later, after done with themes rework (Also will solve confict after 1 more PR)

PR #296, opened 2022-07-15
Allow new mode for Auto ascension to anyone

awWhy:
Allow people at s0 to use new auto Asc mode because that might make game for some people easier to understand
Also small clean up of my toggle PR
(New) prevent being able to use shift + S early (Because spoilers)


awWhy:
Moving color related changes into another PR

PR #298, opened 2022-07-15
fix for auto hept giving free expansions

ab3327:
The initial expand was doing the math for an expansion without increasing the amount crafted(used to subtract hepts from total) and without decreasing the amountToCraft. In the final expansion you were getting a free half an expansion+whatever was left in amountToCraft


ab3327:
The initial expand was doing the math for an expansion without increasing the amount crafted(used to subtract hepts from total) and without decreasing the amountToCraft. In the final expansion you were getting a free half an expansion+whatever was left in amountToCraft

KhafraDev (approved):

PR #301, opened 2022-07-18
New buttons for Orbs: Buy Max and Auto Buy

jarboo:
Added 2 buttons for Orbs.
Auto is handled the same way as Hept crafts.
Visual changes were made to fit those buttons and so that Orbs and Warp display is more consistent with the rest of the page.
New visual for Singularity 1+
image
New visual for SIngularity 0
image

As a reminder, current visual
image


awWhy:
Please add this message
image

awWhy:
Also most likely can reuse rules for .hepteractType
Instead of have extra one

PR #302, opened 2022-07-19
Themes rework

awWhy:
I tryed to remove as much inline css as I can for themes
I'm going to take a break from themes, after some time going to rework light mode (Maybe) and
Add new classes for text colors. As well was planing 1 new theme (Lighest)

### DIFF
--- a/Synergism.css
+++ b/Synergism.css
@@ -18,33 +18,45 @@ html {
 body {
     transform-origin: 0 0;
     overflow-x: hidden;
-    color: white;
     margin: 0;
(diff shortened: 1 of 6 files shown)


KhafraDev (approved):

awWhy:
NOT IN THIS PR:
I made this
image
Also made that themes "save" on refresh (They just get re applyed after game loads)
Wont post it in this PR, to keep this one clean
Because new one might be a bit of Spaghetti, so will wait till this one merged

PR #304, opened 2022-07-22
Visuals

awWhy:
Small stuff I asked for last PR:
Confirmation message before your Hepts go thanos
Make add x1 button look a little better
Noise fix for Bigger orb and powder image
Small stuff from my last PR's:
Properly hide sac button
Fix colors for auto Ascension

Re use some old rules in css and remove unused line
This PR doesnt contain fix for auto orbs not giving powder (?)


src/Hepteracts.ts
```diff
@@ -489,6 +489,12 @@ export const tradeHepteractToOverfluxOrb = async (buyMax?:boolean) => {
     let toUse: number;
 
     if (buyMax) {
+        if (player.toggles[35]) {
+            const craftYesPlz = await Confirm(`This will attempt to buy as many orbs as possible. \nYou can buy up to ${format(maxBuy, 0, true)} with your hepteracts. Are you sure?`);
```
KhafraDev, line 493: I don't think new lines work in here

KhafraDev (approved):

src/Hepteracts.ts
awWhy: It actually does, if you have toggle on on, instead of going to next line it will stop right on alert
I was actually surprised that it does work, if to be honest

PR #306, opened 2022-07-23
Theme rework part 2

awWhy:
Better colors for Dracula theme
4 new custom colors
Window to select themes out of order
Themes get re applied on refresh or import

I will now work on text colors and light mode
Yes it need to be absolute positioned, its a pop up window, you dont want it to push anything around
(Not fixed position, because it need to be affected by scrolling)

(Pushed to clean up a little and to use local storage instead)


Synergism.css
```diff
@@ -741,10 +749,33 @@ nav li:hover {
...
+    border-color: var(--boxmain-bordercolor);
+    margin-left: 10px;
+}
+
+.themeWindow:hover,
+#theme:focus + .themeWindow {
+    display: inline-grid;
+    grid-template-columns: repeat(2, 150px);
+    padding: 2px;
+    position: absolute;
```
KhafraDev, line 776: absolute positioned

src/CheckVariables.ts
```diff
@@ -663,4 +664,11 @@ export const checkVariablesOnLoad = (data: PlayerSave) => {
     if (data.ascensionCounterRealReal === undefined) {
         player.ascensionCounterRealReal = 0;
     }
+
+    //Will re apply theme's on refresh
+    if (data.theme === undefined) {
```
KhafraDev, line 669: this shouldn't be saved on the player object at all

I would just use localStorage since it's not gonna change often

KhafraDev (changes requested):

src/CheckVariables.ts
awWhy: Its only adding 1 number into data

Synergism.css
awWhy: Yes because its a pop up
Same as if you would click your right button on mouse
(That one is absolute positioned as well)

src/CheckVariables.ts
awWhy: Dont want to use localStorage to allow imports to keep theme that you had

awWhy: Also its where its only on load or import is checked and loaded

awWhy: Also if you excepted me to move every color value into local storage, my answer is no

awWhy:
I found some duplicate lines, will clean up them, when will be done with light mode

awWhy:
Closed since last PR, added into player theme value, and I'm not in the mood to work around it

PR #307, opened 2022-07-23
Fix auto Orbs, add text and some noise fixes

awWhy:
Because too many people complain about it
image
There is also a bug with auto being on OFF after import, but showing itself as ON
(New) from cloudy fix wrong text with using adds; from ab33 move TQE below Singularity one
Also fixed a dot that was hiden away from everyone  image


KhafraDev (approved):

cloudytheconqueror:
Previous pull requests related to the add code text fix: #300 #303

awWhy:
OctAsc
Octeract Quarks
I also still have these 2

awWhy:
Removed check for Cookie 3 and Cookie 4 cost on load, because higher in the code there is already better version that does same thing, but for every upgrade (Dont know if it will work properly for more than 1 level, but it works fine for 1 level)

awWhy:
I dont like not being able to squish because of Merge stuff...
Fixed orbs not correctly showing which mode they are after refresh (No idea if in optimal way)

PR #313, opened 2022-07-28
feat: improve offline progress

KhafraDev:
image

Co-authored-by: httpsnet <ituminakao@yahoo.co.jp>


httpsnet:
I think it's a good change
We continue to update # 295, but we plan to create a new PR as the local repository is no longer applicable.

That email address should not be published
It's used for account registration and I never reply

KhafraDev:
I would recommend updating your github email settings (https://github.com/settings/emails) as it is currently public in all of your commits.

PR #314, opened 2022-07-29
Add and settings bugs

mustache0110:
A couple small bug fixes. Headers for special actions and theme are hidden until first prestige while the buttons are always available. I made them always visible.

Renamed droughEffect to the most likely intended droughtEffect. A pedantic clean-up... No logic changed.

And a re-fix of the Add alert for a maxed calculator that was partially reverted with the new 'Add 1' feature. Additionally, changed the alert text so that the maxed and unmaxed calculators are more consistent.


awWhy:
Removing Classes from some settings buttons is already in my themes PR
Fixing add text is already in my auto powder fix
But they shoudnt create any conflicts, as I did they same way

Pseudonian (approved):

PR #316, opened 2022-07-29
Hide toggle

awWhy:
Just a small optimization to that toggle
And hide Ascension upgrades past s51 if maxed

Also I wonder if there easier way to turn them unrefundable
I tryed if (shopData[key].tier === 'Ascension') { shopData[key].refundable = false }
But it didnt worked


awWhy:
Closed. Dont want to deal with conflicts

PR #317, opened 2022-07-30
time to gain 1 oct changes to octs/s at 1 second per oct

ab3327:
When you gain 1 or more octeracts per second it will now display as how many you gain per second rather than how many seconds until you gain 1.


src/UpdateVisuals.ts
```diff
@@ -542,7 +542,9 @@ export const visualUpdateOcteracts = () => {
         return
     }
     DOMCacheGetOrSet('singOcts').textContent = format(player.wowOcteracts, 2, true, true, true)
-    DOMCacheGetOrSet('sPO').textContent = format(1 / octeractGainPerSecond(), 2, true)
+    DOMCacheGetOrSet('sPO0').textContent = (octeractGainPerSecond() < 1) ? '+1 per' : 'You gain'
```
KhafraDev, line 545: +1 every

KhafraDev (changes requested):

src/UpdateVisuals.ts
ab3327: Changed, guess that's what i get for following Plat's guidline

httpsnet:

PR #325, opened 2022-08-04
Fix auto orbs causing hepts to go negative

ab3327:
Two people reported experiencing their cubes going NaN, we were able to use KuWaii's save for narrowing it down to hepts going negative when auto orbs were on and at least one auto hept was on. This seems to happen rarely except on their saves. This save is a prime example of it just ascending once or twice and everything will go NaN after ascending with negative hepts.
[NaN_on_Ascend_sometimes.txt](https://github.com/Pseudo-Corp/SynergismOfficial/files/9262937/NaN_on_Ascend_sometimes.txt)
The fix was just checking if the number of orbs we want to craft would cause a negative amount of hepts and if it would don't craft any.


KhafraDev (approved):

httpsnet:

PR #331, opened 2022-08-09
Preserve autoOrb state across sing

ab3327:
Save auto orb on/off setting on sing


httpsnet:

PR #332, opened 2022-08-10
bug fix for C15 resetting

mustache0110:
I believe this was the intended logic

### DIFF
--- a/src/Synergism.ts
+++ b/src/Synergism.ts
@@ -3028,11 +3028,16 @@ export const resetCheck = async (i: resetNames, manual = true, leaving = false):
             }
         }
 
-        if (a === 15 && (manual || leaving || player.shopUpgrades.challenge15Auto > 0) && player.usedCorruptions.slice(2, 10).every((a) => a === 11)) {
+        if (a === 15 && player.usedCorruptions.slice(2, 10).every((a) => a === 11)) {
             if (player.coins.gte(Decimal.pow(10, player.challenge15Exponent / challenge15ScoreMultiplier()))) {
                 player.challenge15Exponent = Decimal.log(player.coins.add(1), 10) * challenge15ScoreMultiplier();
                 c15RewardUpdate();
             }
+            if (manual) {
+                player.currentChallenge.ascension = 0;
+                updateChallengeDisplay();
+            }
+            return;
         }
 
         if (player.challengecompletions[a] > player.highestchallengecompletions[a]) {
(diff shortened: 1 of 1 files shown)


KhafraDev (changes requested):
This doesn't seem to fix entering c15 w/ auto sweep enabled and retry challenges disabled(?).

mustache0110:
Honestly, still reading through the changes... and I had assumed it was the addition of that shop item that caused the change in logic to calling a reset on C15 which can never reset. I thought that the intent was to use the autoC15 to update C15exponent. That's what I get for trying it digest a bunch of code while tired, I totally missed the expansion of auto-challenge above sing 101 :(

I'm not sure what the intended behavior is. Are we putting C15 into reset for both the new shop upgrade and auto-challenge expansion

mustache0110:
```        if (!player.retrychallenges || manual || leaving) {
            if (!(!manual && (autoAscensionChallengeSweepUnlock() || !player.autoChallengeRunning) // If not autochallenge, don't reset
                             && player.autoAscend && player.challengecompletions[11] > 0
                             && player.cubeUpgrades[10] > 0)) {
                player.currentChallenge.ascension = 0;
                updateChallengeDisplay();
            }
        }```
Is pretty unclear / unreadable because of the double negatives, at least for my tired brain. player.autoAscend being false would make the entire && chain false, which makes the if statement true and resets our C15. Likewise autoAscensionChallengeSweepUnlock || !player.autoChallengeRunning evaluates to false leading to reset. I'm not sure why challengecompletions[11] is there.

This fix at least reverts the previous behavior for people without Sing > 101 and the shop upgrade.

mustache0110:
> This doesn't seem to fix entering c15 w/ auto sweep enabled and retry challenges disabled(?).

C15 is a special case and the only reason to enter a resetCheck is to update the exponent if the autoC15 shop item is purchased.

The autoascendchallenge logic is implemented in runchallengesweep in Challenges.ts.

I'm seeing plenty of bugs to fix, but this PR is specifically to get C15 working for most people again. I've tested and it works in Sing 1 and 2, and Sing 107 with autoascendchallenge.

It breaks with shop item buyable starting in sing 83, working on a fix now.

KhafraDev (approved):
Love it. Thanks for explaining everything :)

mustache0110:
This fixes the shop upgrade, too. I feel like the logic of the entire section could be cleaned up but I need to step away for a while

mustache0110:
closing since this fix is (thankfully) now redundant and not healthy to merge

PR #327, opened 2022-08-08
Improved updates

httpsnet:
Sorry for not being able to prepare a decentralized PR. I was able to focus on development and solve many problems in a short period of time. appreciate.
I think this PR still has a lot to fix. It saddens me that you deleted my duplicate Auto Open Cube.
I will work for Marge from now on. Please have a discussion.
We apologize for any conflicting contributors.
The update is very long, please pay attention to the sentences due to automatic translation.

Visual renewal of loading and offline progress
Made many layouts em size. This changed the layout size and position. The layout is less likely to break even in environments with large font sizes.
Fixed many issues with offline progress and title screens and backgrounds.
The offline progress button is now automatically focused.
Offline progress is hidden when you first access the game.
Improved alert, confirm, prompt and notification boxes.
Disabled the default display state of each tab. This will improve game loading speed.
Offline progress has gained elasticity.
Offline progress can now be dismissed by double-clicking.
Now closes open notifications at the end of loading.
Changed from creating a function to calling a function inside some interval().
Horizontal scrolling now works.
Adapted the design of the scroll bar to match the game.
I installed a scroll bar for layouts where the text size could exceed 1250px even if the text size is 1em and 16px.
Hotkeys disabled the browser's inferred behavior.
Changed the display of runes and talismans.
Removed some unused CSS. Consolidated some CSS that uses the same elements.
Changed the Toggle amount to buy text to be displayed by title.
Added $STAGE$ to the save data file name identifier. This will show your current position.
Show the reset game button 10 seconds after the page loads. This is useful if the game won't launch for some reason.
Fixed many bugs in Automatically Run Chal.
The display of shop, rune and talisman levels will be hidden under certain conditions.
As a workaround for free singularity, we no longer increment the singularity count if an error prevents the code from running completely.
If Ascension Confirmation Toggles is OFF, removes restrictions on ascension challenge toggles.
So many CSS and style changes.
Cookie upgrade now checks if cookies are unlocked.
Fixed a bug where purchasing Cx7 did not change the background color.
Shop upgrades now check if they are unlocked.
Shop Confirmations: ON, Buy Max: OFF and No refunds no longer confirm purchases.
Decimal for some calculations.
Auto Ascension now works without Ascension Challenge.
Execution of Auto Ascension requires passing Challenge 10.
Challenge Sweep now skips the next challenge even if it's the same challenge as the current one.
Singularity is 2 or higher and changed auto-challenge start index to 10 in start of ascension challenge.
[...]

PR #334, opened 2022-08-10
Fast fix

httpsnet:
We had hoped that Improved updates #327 would be vetted and reviewed, but they were committed without doing so, causing some problems.
I fixed it quickly, and I will continue to fix it. Sorry for the inconvenience.

Golden Quarks I, Golden Quarks II and Golden Quarks III now calculate a free level soft cap.
Fixed C15 score update bug. You may still have problems.
You will no longer exit C15 without purchasing Instant Challenge 2.
C15 can now be completed.
Achievement 252 can now be unlocked.
The maximum number of completions for C15 will be 10. The condition is 1.00e1.000 No Coins. The coin multiplier for each achievement request is x1,000.
C15 may be rewarded in the future.
Fixed Obtainium calculation.
Fixed Cx13 calculation broken when Hepteracts is negative.


src/Challenges.ts
```diff
@@ -52,7 +52,7 @@ export const getMaxChallenges = (i: number) => {
     if (i <= 15 && i > 10) {
         //Challenge 15 has no formal cap, so return 9001.
         if (i === 15) {
-            return 1
+            return 10
```
KhafraDev, line 55: this shouldn't need to be set; should be reverted to 0

KhafraDev (changes requested):

src/Challenges.ts
httpsnet: ok
It took me about an hour to get these working

KhafraDev: this should just handle display, ie sadistic ii used to display 0/0 completions.

PR #338, opened 2022-08-12
fix and update

httpsnet:
Fixed an issue where Achievements 63, 75, 76, 77 and 247 could not be earned.
Fixed the #sacrificeAnts box sometimes overlapping Ant Tier upgrades and not pressing.
Fixed Ascend History and Singularity History layouts.
Updated CSS.
Added $STAGES$. This shows the current stage including Singularity.
Fixed an issue with Last enabled key not working.
Ascension and Ascension Challenges can no longer be completed without Achievements 141. This could work with things like Auto Challenge Sweep and hotkey and HTML editing exploits.
Added Singularity Count to Miscellaneous Stats.

Added Auto Ascend, All Open 10% Cubes, All Open 50% Cubes and All Open All Cubes to Hotkeys.
The Hotkeys list is more visually appealing.
Hotkeys list now shows Description on hover. The text is inadequate.
Numeric hotkeys are also disabled when the dialog is visible.
If the same key is set for hotkey and number hotkey, they will not work at the same time.
If the key setting prompt is open and the textbox is not focused, typing will write the key to the textbox.
Changed EventListener registration for eventHotkeys() to occur at the end of document.onload.

fix and more QoL

Enabled revealStuff() on autoupgrades if you haven't unlocked Ascension.
Singularity upgrades free levels in daily codes are no longer numbered.
Fixed Talisman Shards Auto Buy being able to function without being unlocked.
Updated r3x21-r3x25 effects.
Updated internal code for Toggle amount to buy. No feature additions at this time, but will be in the near future.
Saves the current game version that started the save data to the save data.
The Automations toggle allows you to control which upgrades you get on reset.
Shared shop potions and singularityQuarks elements.
Added Descriptions to singularityQuarks.

Cx1 is now automatically obtained upon ascension from Singularity 11.
Auto Challenge Sweep will be enabled from the start with Singularity 40.
Auto Ascend Mode: Real time reduced to Singularity 5.
Automation Shop is automatically reduced to Singularity 20 and added an Automation Shop toggle.
Autobuy Talismans Resources have been reduced to Singularity 25 and no longer require Cx1.
Added Auto Runes to Hotkeys.


src/Hotkeys.ts
```diff
@@ -176,9 +212,13 @@ const makeSlot = (key: string, descr: string) => {
     });
 
     const p = document.createElement('p');
-    p.id = 'hotKeyDesc';
     p.textContent = descr;
 
+    p.addEventListener('click', () => {
+        DOMCacheGetOrSet('hotkeyDescription').textContent = hotkeys.get(key)?.[3] ?? '';
+        hotkeys.get(key)![1]();
```
KhafraDev, line 219: this logic doesn't make sense on one line you use optional chaining and the next you assert the value cannot be null/undefined

KhafraDev (changes requested):

src/Hotkeys.ts
httpsnet: Didn't need it as pointed out

httpsnet:
Closed because I created a new PR for a similar one

PR #340, opened 2022-08-13
Enhance Quark Shop: Buy {1 | 10 | Max | Any}

mustache0110:
Quark Shop Enhancement:
Took the idea from buy-many-potions and applied it to the Quark Shop in whole.
Changed Buy Max toggle to Buy: {1 | 10 | Max | Any} on all upgrades/potions
Buy 10 feels very nice; Buy ANY is maybe for the more compulsive among us.
Keeps httpsnet's change disabling warning on nonrefundable AND buy 1.

Room for improvement:
Does not honor the singularity perk change to nonrefundable of the various Reincarnation/Ascension level purchases, but they are so cheap at that point that it's meaningless.

It's possible that it's too annoying to click 4 times to cycle through the toggle; but at least clicking 2 times gets you from 1 to Max and Max to 1, which mirrors the old behavior. If there's enough demand it might be enough to drop the cycle to {1 | 10 | Max} when clicking on the left 80% of the button, and have to click on the text indicated ANY section to enter that mode. Mock-up below.
```
___________________
|             | A |
|  Buy: 10    | N |
|_____________| Y |

```
Honestly don't think we spend enough time respecing shop to make the clicks an issue, especially with how much Buy 10 saves. My slight level of annoyance could just be from the amount of time I spent testing in shop at various quark levels.

Changed 'no refund' shop alert to be a little more obvious, as a separate commit in case you want to cherry pick. Although it's been in Unicode since 2010, supported by javascript, and fails gracefully.
⚠️!! No Refunds !!⚠️

### DIFF
--- a/index.html
+++ b/index.html
@@ -3363,12 +3363,12 @@ $STAGE$ for synergism stage" style="color: white"></p>
                             <p id="instantChallenge2Level" alt="instantChallenge2Level" label="instantChallenge2Level"></p>
                             <button class="permshopbutton" id="instantChallenge2Button" alt="instantChallenge2Button" label="instantChallenge2Button" style="border: 2px solid darkviolet"></button>
                         </div>
-                        <div class="chal8Shop" id="antSpeedHide">
+                        <div class="chal8" id="antSpeedHide">
                             <img id="antSpeed" alt="antSpeed" label="antSpeed" src="Pictures/antspeed.png" loading="lazy">
                             <p id="antSpeedLevel" alt="antSpeedLevel" label="antSpeedLevel">Level: 0/3</p>
                             <button class="permshopbutton" id="antSpeedButton" alt="antSpeedButton" label="antSpeedButton" style="border: 2px solid maroon">Upgrade for 200 Quarks</button>
                         </div>
-                        <div class="chal8Shop" id="cashGrabHide">
+                        <div class="chal8" id="cashGrabHide">
                             <img id="cashGrab" alt="cashGrab" label="cashGrab" src="Pictures/cashgrab.png" loading="lazy">
                             <p id="cashGrabLevel" alt="cashGrabLevel" label="cashGrabLevel">Level: 0/5</p>
                             <button class="permshopbutton" id="cashGrabButton" alt="cashGrabButton" label="cashGrabButton" style="border: 2px solid limegreen">Upgrade for 100 Quarks</button>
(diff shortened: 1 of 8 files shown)


mustache0110:
After some thought, I think toggling between 1, 10, MAX makes the most sense. With a Shift-click on the toggle to switch to Buy: ANY. My pause is how to alert the user.

Mouse-over text via title="Shift-click to toggle Buy: Any" 
vs
The same text on the button as a second line. Would want it to be smaller to not pull focus from the toggle status.

mustache0110:
@KhafraDev @Pseudonian Ready for review / comments / questions. I rebased on the current master and put a test version up at https://mustache0110.github.io/SynergismOfficial/ if you want to see how the shop feels. I think it feels great without that extra click in the cycle.

KhafraDev (approved):

mustache0110:
Tacked these shop changes onto the old PR. Equi-buy depends on Enhance Quark Shop changes.

> Room for improvement:
> Does not honor the singularity perk change to nonrefundable of the various Reincarnation/Ascension level purchases, but they are so cheap at that point that it's meaningless.

Got it.

The perk fixes should be bundled and conflict with #357

awWhy:
I would recommend:
1. Replace all 'shopData[key]' with 'shopItem'  in UpdateVisuals.ts from line 593 to 637
2. Remove const like perk_5, that is used only once and just use achievement check directly (really no reason not to, can add comment, if something)
3. Probably all special classes for unlock can be removed from HTML for all shopUpgrades, since there were only used in hide toggle

mustache0110:
> I would recommend:
> 
>     1. Replace all 'shopData[key]' with 'shopItem'  in UpdateVisuals.ts from line 593 to 637
> 
Good call, I had changed a bunch and missed a few.

>     2. Remove const like perk_5, that is used only once and just use achievement check directly (really no reason not to, can add comment, if something)
That boolean is used for the ternary operator below, so i figured why not move it up and use it one more time.
 
>     3. Probably all special classes for unlock can be removed from HTML for all shopUpgrades, since there were only used in hide toggle
I'm pretty sure they're used in revealStuff() and should not be touched. 3382cd is a relevant commit and while they could be removed/reverted, it would also require some code changes beyond removing the classes. I'll look into it.

awWhy:
> I'm pretty sure they're used in revealStuff() and should not be touched. 3382cd is a relevant commit and while they could be removed/reverted, it would also require some code changes beyond removing the classes. I'll look into it.

image
Well all these classes were added by me specifically for hide max toggle, if all logic were erased of that toggle, then these classes don't actually do anything (if nothing changed)

mustache0110:
They're still required in some form because you do need to differentiate between chal14 unlocks (researches / achievements / whatever) and the shop unlocks,specifically. Otherwise DOMCacheGetOrSet(`${key}) hides everything when you only want the shop items to be affected. Because the visibility is already handle in shop display code, there's no need to revert them to back to work with revealStuff().

The toggle logic was not removed, just trimmed to 5 lines. (From 100+!, using the function Platonic added for Summary update)

mustache0110:
Just as a side-note: I had this helper function before I found the other one (which obviously didn't exist when you added the feature):
```
+    const shopUnlocked = (key: keyof Player['shopUpgrades']) => {
+        const shopClassless = ['offeringAuto', 'offeringEX', 'obtainiumAuto', 'obtainiumEX','instantChallange', 'calculator'];
+        if (shopClassless.includes(key)) { return true; }
+
+        const shopClass = DOMCacheGetOrSet(`${key}Hide`).className;
+        if (shopClass === 'chal8Shop') { return player.achievements[127] > 0; }
+        if (shopClass === 'chal9Shop') { return player.achievements[134] > 0; }
+        if (shopClass === 'ascendunlockShop') { return player.ascensionCount > 0; }
+        if (shopClass === 'chal11Shop') { return player.challengecompletions[11] > 0; }
+        if (shopClass === 'chal12Shop') { return player.challengecompletions[12] > 0; }
+        if (shopClass === 'chal14Shop') { return player.challengecompletions[13] > 0; }
+        if (shopClass === 'chal13Shop') { return player.challengecompletions[14] > 0; }
+        if (shopClass === 'hepteractsShop') { return player.challenge15Exponent >= 1e15; }
+        if (shopClass === 'singularityShopUnlock') { return player.singularityUpgrades.wowPass.getEffect().bonus; }
+        if (shopClass === 'singularityShopUnlock2') { return player.singularityUpgrades.wowPass2.getEffect().bonus; }
+        if (shopClass === 'singularityShopUnlock3') { return player.singularityUpgrades.wowPass3.getEffect().bonus; }
+    }
```
I felt good because it let me clean up a lot of lines and only run the check once on the key in the loop, rather than all keys on every key loop. But then the linter complained only one statement per line, and I felt bad. I realized switch(shopClass) works better, and rewrote it that way. I felt bad I didn't think of it first, but also good because it's better. Finally, I found the summary function and felt bad again, as I deleted my helper. A roller-coaster of emotions.

awWhy:
Um, yes lint by default has only 1 statement per line, in my code I allowed for 2 min for casses like these. But that whole block was better to be done in switch case for better perfomance (or at least else if chain);
>The toggle logic was not removed, just trimmed to 5 lines. (From 100+!, using the function Platonic added for Summary update)

That logic was also all the code that was used for revealing stuff (with is why it was so long). So revealStuff() shoudn't use any of these classes. Only GQ buying wasn't done with that code. These classes most likely can be removed safely, just search in vscode for these classes names

awWhy:
By the way, this is not what I asked, I asked to remove classes not to rename them to ne managed with revealStuff.
Because hide toggle took work of a revealStuff for entire shop, changing class name into these one's will only do useless checks

PR #342, opened 2022-08-14
Fix autoTess amount incorrectly showing

ab3327:
This PR prevents the text box from showing a different value than the text in percent mode. Currently, when you enter a number > 100 the text below the box corrects to 100% while the text in the box stays at the entered value. This PR will set the text in the box to 100 if the text entered is >100 and it is toggled to % mode. This will help prevent confusion when people swap between % and amount modes.
This image shows the text as 100% but the box shows 1e3.
image

### DIFF
--- a/src/Reset.ts
+++ b/src/Reset.ts
@@ -173,6 +173,9 @@ export const updateAutoReset = (i: number) => {
 export const updateTesseractAutoBuyAmount = () => {
     let value = Math.floor(parseFloat((DOMCacheGetOrSet('tesseractAmount') as HTMLInputElement).value)) || 0;
     if (player.resettoggle4 === 2) { // Auto mode: PERCENTAGE
+        if (value > 100) {
+            (DOMCacheGetOrSet('tesseractAmount') as HTMLInputElement).value = '100';
+        }
         value = Math.min(value, 100);
     }
     player.tesseractAutoBuyerAmount = Math.max(value, 0);


KhafraDev (changes requested):
I think I have a better approach. Blocking so @Pseudonian doesn't merge.

mustache0110:
```
diff --git a/src/Reset.ts b/src/Reset.ts
index 5b5bfde2..9a7f08ef 100644
--- a/src/Reset.ts
+++ b/src/Reset.ts
@@ -171,10 +171,7 @@ export const updateAutoReset = (i: number) => {
 }
 
 export const updateTesseractAutoBuyAmount = () => {
-    let value = Math.floor(parseFloat((DOMCacheGetOrSet('tesseractAmount') as HTMLInputElement).value)) || 0;
-    if (player.resettoggle4 === 2) { // Auto mode: PERCENTAGE
-        value = Math.min(value, 100);
-    }
+    const value = Math.floor(parseFloat((DOMCacheGetOrSet('tesseractAmount') as HTMLInputElement).value)) || 0;
     player.tesseractAutoBuyerAmount = Math.max(value, 0);
 }
```
Just putting this here in case ab33 doesn't see my comment in the bug report thread:
> I think there's a better fix because the way it is now still makes it pretty easy for a user to spend all tesseract cubes. There's no reason to clobber the value on a toggle change, which can easily happen. It's sandwiched between the two fields they probably use most often.
> 
> I think the bug is that hitting the amount/percent toggle changes the value to 100 when it should leave it alone.
> 
> if you think about it, changing the value has NO EFFECT on someone who is buying by percentage, but screws over someone early in the game where tesseracts are hard to come by.
> 
> Not changing the value also has NO EFFECT on someone buying by percent, and does not screw over the early game player.

With the suggested change, the calculating code and display code already work with no additional changes because they already use a Math.min(100, value) on percentage buy.

PR #356, opened 2022-09-12
Summary firstplayed

mustache0110:
easiest way to fix reported bug with summary by RandomFella. 

also noticed export not being entirely truthful, once you exported a file it permanently said Exported to file!. I can almost guarantee that no one was confused by this because they never read it, but that doesn't mean we can allow the program to lie to us.

### DIFF
--- a/src/ImportExport.ts
+++ b/src/ImportExport.ts
@@ -149,6 +149,7 @@ export const exportSynergism = async () => {
             // - TypeError (browser doesn't support this feature)
             // - Failed to copy (browser limitation; Safari)
             await navigator.clipboard.writeText(saveString)
+            DOMCacheGetOrSet('exportinfo').textContent = 'Copied save to clipboard!';
         } catch (err) {
             // So we fallback to the deprecated way of doing it,
             // which isn't limited by any browser.
@@ -177,6 +178,7 @@ export const exportSynergism = async () => {
             })
 
             clipboard.on('error', () => {
+                DOMCacheGetOrSet('exportinfo').textContent = 'Export failed! DM Ant God'
                 void Alert('Unable to write the save to clipboard.').finally(cleanup)
             })
         }
@@ -191,9 +193,9 @@ export const exportSynergism = async () => {
         document.body.appendChild(a);
         a.click();
         document.body.removeChild(a);
+        DOMCacheGetOrSet('exportinfo').textContent = 'Savefile copied to file!';
     }
-
-    DOMCacheGetOrSet('exportinfo').textContent = 'Savefile copied to file!';
+    setTimeout(() => DOMCacheGetOrSet('exportinfo').textContent = '', 15_000);
 }
 
 export const reloadDeleteGame = async () => {
--- a/src/ImportExport.ts
+++ b/src/ImportExport.ts
@@ -178,7 +178,7 @@ export const exportSynergism = async () => {
             })
 
             clipboard.on('error', () => {
-                DOMCacheGetOrSet('exportinfo').textContent = 'Export failed! DM Ant God'
+                DOMCacheGetOrSet('exportinfo').textContent = 'Export failed!'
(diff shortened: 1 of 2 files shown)


mustache0110:
Thinking about it while driving home, and various use cases...

A) sharing saves to help
B) different people using same computer with different progress
C) people starting over/ challenge runs / speed runs
D) same save on different computers

I guess you do want to attach the firstPlayed to the save file itself only. The issue being that what do you do on files missing the data? Is it ok to just ignore because if the user truly cares enough they can hand edit the file. Or do you want to add a way to 'import oldness' in the stats for nerds panel where they can input an old save that will update the current save's lastPlayed to the dateModified on the validated input save. This seems like the way to go, IMO.

httpsnet:
Looking at the code, everything seems correct.
I think that a function to import the save data acquired by Copy to Clipboard as text should be implemented.
I've lost my credibility, so I'm limited in what I can do.

src/ImportExport.ts
```diff
@@ -177,6 +178,7 @@ export const exportSynergism = async () => {
             })
 
             clipboard.on('error', () => {
+                DOMCacheGetOrSet('exportinfo').textContent = 'Export failed! DM Ant God'
```
KhafraDev, line 181:
```suggestion
                DOMCacheGetOrSet('exportinfo').textContent = 'Export failed!'
```

Errors will likely only come from mobile safari, which limits how copy/paste works. Nothing we can do if it occurs, so no point in having someone dm @Pseudonian over this

KhafraDev (approved):

KhafraDev:
@httpsnet You haven't lost your credibility, at least neither plat or I think so.

src/ImportExport.ts
mustache0110: Was just being cheeky. Maybe should have been Pray to the Ant God. Or Prey of the Ant God. But fair point, would be smart to avoid actual PMs

mustache0110:
> Looking at the code, everything seems correct. I think that a function to import the save data acquired by Copy to Clipboard as text should be implemented. I've lost my credibility, so I'm limited in what I can do.

First of all, everyone makes mistakes. Me, you, Khafra, even JavaScript. Bugs are unavoidable. So there is no shame in that. Truly. And if we're talking about PR#327, there was no single mistake made. Translation might have been an issue too. I think you said it needed testing but the translation of "I think this PR still has a lot to fix" did not properly convey that. I know it didn't to me until i re-read it later. You are able to create a pull request as a draft, which means it cannot be merged. That was your only real mistake, which is pretty small.

I also almost jumped in to suggest putting this on the test server, but I hesitated because I felt my level was too low. So I definitely share that blame. But we all learn things every day, and that was my mistake.

Second, it's hard to read someone else's code, especially if it's not well documented, because you are not sure of their intent. You don't know what their assumptions are and if they're well founded or overlooking something. It's harder to have confidence when reviewing because of this, which is why I would suggest smaller commits. They can all go into the same branch, but it's easier to review smaller commits. CSS changes all in one commit, white-space or formatting in one commit, refactoring, function change, new features all in separate commits. Try to make it easier for the reviewer because teamwork.

Third, remember that this project is managed by Plat and Khafra in their spare time around other obligations. Another reason to try to make it easier for them, and not expect them to be able to respond immediately.

Lastly, if you really want to think about your credibility... Just look at the results. They did not revert your commit. Everyone is playing on it. Bugs were fixed, and the new features you added are here, too. Like I said on discord, "I do think your ideas and effort are appreciated. You have many great ideas"

Your English is very good and it's hard to remember sometimes on the internet that not everyone is a native speaker. I really hope you do continue to contribute. Hopefully you read this with the positivity I intended. My apologies if this isn't the right place for this, but it's more secluded here than discord. Because secluded is totally a word you should use with a non-native speaker. But when I'm tired I just don't make sense. Hence, Pray to the Ant God.

PR #357, opened 2022-09-18
Reopening old PR for hide toggle

awWhy:
Small optimization, remove no longer usefull check
Better perfomance will be if to just write for each key: refundable = false; But this way is easier to add extra's
4x 'as keyof typeof shopData' is auto removed when converted to js, so doesn't matter if ugly

### DIFF
--- a/src/CheckVariables.ts
+++ b/src/CheckVariables.ts
@@ -231,6 +231,31 @@ export const checkVariablesOnLoad = (data: PlayerSave) => {
         player.shopUpgrades.tesseractToQuark = 0;
         player.shopUpgrades.hypercubeToQuark = 0;
     }
+    for (const i in shopData) { //Since you cant increase SingularityCount without calling it
+        if (shopData[i as keyof typeof shopData].tier === 'Reincarnation') {
+            if (player.singularityCount >= 20) {
+                shopData[i as keyof typeof shopData].refundable = false;
+            } else {
+                shopData.offeringAuto.refundable = true;
+                shopData.offeringEX.refundable = true;
+                shopData.obtainiumAuto.refundable = true;
+                shopData.obtainiumEX.refundable = true;
+                shopData.antSpeed.refundable = true;
+                shopData.cashGrab.refundable = true;
+            }
+        } else if (shopData[i as keyof typeof shopData].tier === 'Ascension') {
+            if (player.singularityCount >= 51) {
+                shopData[i as keyof typeof shopData].refundable = false;
+            } else {
+                shopData.seasonPass.refundable = true;
+                shopData.seasonPass2.refundable = true;
+                shopData.seasonPass3.refundable = true;
+                shopData.seasonPassY.refundable = true;
+                shopData.chronometer.refundable = true;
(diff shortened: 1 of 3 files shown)


src/CheckVariables.ts
```diff
@@ -231,6 +231,13 @@ export const checkVariablesOnLoad = (data: PlayerSave) => {
         player.shopUpgrades.tesseractToQuark = 0;
         player.shopUpgrades.hypercubeToQuark = 0;
     }
+    for (const i in shopData) { //Since you cant increase SingularityCount without calling it
+        if (shopData[i as keyof typeof shopData].tier === 'Reincarnation') {
+            shopData[i as keyof typeof shopData].refundable = player.singularityCount >= 20 ? false : true;
```
mustache0110, line 236: Would this make all shop items refundable below 20? Including instantChallenge, and shopTalisman, (and potions). I think you mean to only change the ones where refundable === true.

I like it because it makes hide maxed work better. I think it belongs in the updateSingularityMilestoneAwards in Reset.ts, though.

awWhy: Oh wait that probably true, will deal with it shortly
About putting it in reset, don't really want because it has to happen on refresh and preferably only once per reload
It can be moved anywhere into reload function, but I don't know how often it is called and when

awWhy:
If there was something like blanksave.shopdata then could make it more optimized by adding 'blanlsave.shopdata[i].refundable === false'
Code can be moved out of checkVariable, but preferably into where it will be called only once on save load

mustache0110:
> Code can be moved out of checkVariable, but preferably into where it will be called only once on save load

So, to me, it makes sense to put it into updateSingularityMilestoneAwards() for 2 reasons.

1. checkVariablesOnLoad() is pretty specific about its role if you read the documentation at the top
/**
 \* Given player data, it checks, on load if variables are undefined
 \* or set incorrectly, and corrects it. This should be where all new
 \* variable declarations for `player` should go!
 \* @param data
 \*/
So it's more for fixing bugs / undefineds and it's confusing to start handling perks in checkVariablesOnLoad() as well.

2. As for updateSingularityMilestoneAwards(), many/most other perks are handled there and that's the first place I would look if debugging code. It very much is a perk milestone, very much feels like it belongs here, and there would have to be a very good reason to put it elsewhere (IMO).  

3. If you only want it to run once in that code, you can use something like if (player.singularityCount >= 20 && singularityReset). A singularity reset calls the milestoneawards, while other resets, i.e. challenges/transcend/reincarnate/ascension call milestoneawards(false).

4. Optimizing is good and a good thing to think about. But as compilers get better, code readability and maintainability are more important than micro-optimization. And a profiler is a good tool to see if it's worth your time. If this function runs less than 1% of the time and you can triple its speed, you won't even notice.

awWhy:
updateSingularityMilestoneAwards, i'm pretty sure has some drawbacks, its not always called on reload.
I'll let plat to decide where to move it, since it really doesn't matter

awWhy:
#340 
Should be doing same thing, but probably better
No idea if there any bugs inside
So this PR is closed

PR #358, opened 2022-09-20
update interval calculation to include PL-AT Σ perk

mustache0110:
Also added the stated cap of 50%, which wouldn't be realized until Singularity 250?

### DIFF
--- a/Synergism.css
+++ b/Synergism.css
@@ -2161,6 +2161,7 @@ p#reincarnatehotkeys {
     min-width: 100px;
     color: white;
     background-color: black;
+    z-index: 1;
 }
 
 p[id$="BlessingsTotal"] {
--- a/src/ImportExport.ts
+++ b/src/ImportExport.ts
@@ -663,6 +663,8 @@ const addCodeMaxUses = () : number => {
 const addCodeInterval = () : number => {
     let time = hour
     time *= (1 - 0.02 * player.shopUpgrades.calculator4)
+    time *= (1 - Math.min(.5, (player.highestSingularityCount >= 125 ? player.highestSingularityCount / 1000 : 0)
+                            - (player.highestSingularityCount >= 200 ? player.highestSingularityCount / 1000 : 0)))
     return time
 }
 
@@ -674,12 +676,7 @@ const addCodeAvailableUses = (): number => {
 }
 
 const addCodeTimeToNextUse = (): number => {
-    let timeInterval = hour
-    timeInterval *= (1 - 0.02 * player.shopUpgrades.calculator4)
-    timeInterval *= (1 - (player.highestSingularityCount >= 125 ? player.highestSingularityCount / 1000 : 0)
-                       - (player.highestSingularityCount >= 200 ? player.highestSingularityCount / 1000 : 0))
-
-    return Math.floor(timeInterval + player.rngCode - Date.now())/1000;
+    return Math.floor(addCodeInterval() + player.rngCode - Date.now())/1000;
 }
 
 const timeCodeAvailableUses = (): number => {
--- a/src/Synergism.ts
+++ b/src/Synergism.ts
@@ -858,6 +858,9 @@ const loadSynergy = async () => {
         Object.defineProperty(window, 'Decimal', {
             value: Decimal
         });
+        if (data) {
+            data.exporttest = false;
+        }
     }
 
     Object.assign(G, { ...blankGlobals });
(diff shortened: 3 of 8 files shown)


mustache0110:
Issue reported by RandomDude, https://discord.com/channels/677271830838640680/1021203206019235860/1021203208258998412

Was able to reproduce by using an add code as soon as it became available, resulting in a negative timer which increases. Due to timeInterval discrepancy, the display code uses the Perk while the use code does not and sets player.rngCode a little too far in the future. As the perk grows stronger it becomes more noticeable.

KhafraDev (approved):
thanks

Synergism.css
```diff
@@ -2161,6 +2161,7 @@ p#reincarnatehotkeys {
     min-width: 100px;
     color: white;
     background-color: black;
+    z-index: 1;
```
mustache0110, line 2164: godskook's fix for platonic cube auto-open bug (placed under other elements)

src/ImportExport.ts
```diff
@@ -674,12 +676,7 @@ const addCodeAvailableUses = (): number => {
 }
 
 const addCodeTimeToNextUse = (): number => {
-    let timeInterval = hour
-    timeInterval *= (1 - 0.02 * player.shopUpgrades.calculator4)
-    timeInterval *= (1 - (player.highestSingularityCount >= 125 ? player.highestSingularityCount / 1000 : 0)
-                       - (player.highestSingularityCount >= 200 ? player.highestSingularityCount / 1000 : 0))
-
-    return Math.floor(timeInterval + player.rngCode - Date.now())/1000;
+    return Math.floor(addCodeInterval() + player.rngCode - Date.now())/1000;
```
mustache0110, line 679: including awWhy's better bug fix, this way there aren't two places to update the interval calculation (which caused the bug initially)

src/Synergism.ts
```diff
@@ -4011,8 +4014,11 @@ export const reloadShit = async (reset = false) => {
 
 window.addEventListener('load', () => {
     const ver = DOMCacheGetOrSet('versionnumber');
+    const addZero = (n: number) => {
+        return (n<10 ? '0' + n : n);
+    }
     if (ver instanceof HTMLElement) {
-        const textUpdate = !isNaN(lastUpdated.getTime()) ? ` [Last Update: ${lastUpdated.getHours()}:${lastUpdated.getMinutes()} UTC ${lastUpdated.getDate()}-${lastUpdated.toLocaleString('en-us', {month: 'short'})}-${lastUpdated.getFullYear()}].` : '';
+        const textUpdate = !isNaN(lastUpdated.getTime()) ? ` [Last Update: ${addZero(lastUpdated.getHours())}:${addZero(lastUpdated.getMinutes())} UTC ${addZero(lastUpdated.getDate())}-${lastUpdated.toLocaleString('en-us', {month: 'short'})}-${lastUpdated.getFullYear()}].` : '';
```
mustache0110, line 4021: Surprisingly(?) this is the easiest way to handle in javascript. fixes the bug of time showing up as 18:5 instead of 18:05, and if we're using 24-hour clock then technically we want a leading zero on the hour too. Why not also include the day.

src/Synergism.ts
```diff
@@ -858,6 +858,9 @@ const loadSynergy = async () => {
         Object.defineProperty(window, 'Decimal', {
             value: Decimal
         });
+        if (data) {
+            data.exporttest = false;
```
mustache0110, line 862: If we're attaching the variables to window, we want to mark the save as a testing save. Otherwise it inherits the value from the loaded save file.

mustache0110 (commented):
annotated the non-spelling changes

src/Synergism.ts
```diff
@@ -4011,8 +4014,11 @@ export const reloadShit = async (reset = false) => {
 
 window.addEventListener('load', () => {
     const ver = DOMCacheGetOrSet('versionnumber');
+    const addZero = (n: number) => {
+        return (n<10 ? '0' + n : n);
+    }
```
KhafraDev, line 4019:
```suggestion
    const addZero = (n: number) => `${n}`.padStart(2, '0')
```

KhafraDev (changes requested):

src/ImportExport.ts
awWhy: Sadly not mine fix, but httpnet one
(I think)

mustache0110: you are correct, i think i was talking to you about another issue when I saw his PR and confused the two >_<

PR #359, opened 2022-09-20
fix add message

httpsnet:
fix the bug that the number of seconds of the next add is negative


KhafraDev:
is this a duplicate of #358?

httpsnet:
It seems that it overlapped for several tens of minutes, I think that I am an unnecessary person

PR #365, opened 2022-09-30
Fix for Singularity History

Mixelz:
The scrollbarX Div class was forgotten about on the Singularity History page. 
Adding it back in in fixes the misalignment issue that page has had since its inclusion. 
This is a very minor change to the HTML file, I believe merge conflicts should be the only primary concern here?
Hopefully so anyways.


KhafraDev:
Thank you for your contribution ❤️

PR #367, opened 2022-10-02
Penalty Difficulty Indicators

Mixelz:
Added a few indicators to the Penalty tab to indicate difficulty spikes! 
In specific, there are indicators for: 
When a new Penalty will be unlocked.
When your effective singularity count will be sharply increased.

These changes are meant to help players more easily understand when difficulty spikes are to be expected so that they can better prepare for them and not get blindsided by them unexpectedly.

New penalty unlocks are currently indicated via the following message, replacing the current empty space with:
"???????? ??????? ????? ??? ?????????? ?? ??? (Sing X)" 
This can easily be replaced with something like "A new penalty unlocks in Singularity X" if it is so wished.

The penalty difficulty spikes are indicated via an additional line of text that reads:
"Your penalties will sharply increase in Singularity X"
I am not a good programmer so I believe this could be more elegantly implemented but my baseline can be worked off of. 

Example Image of these changes in action:

unknown (4)

### DIFF
--- a/src/singularity.ts
+++ b/src/singularity.ts
@@ -1475,7 +1475,7 @@ export const calculateEffectiveSingularities = (singularityCount: number = playe
 export const calculateNextSpike = (singularityCount: number = player.singularityCount): number => {
     const singularityPenaltyThreshold = [11, 26, 37, 51, 101, 151, 250];
     for (const sing of singularityPenaltyThreshold) {
-        if (sing >= singularityCount) {
+        if (sing > singularityCount) {
             return sing;
         }
     }
--- a/src/singularity.ts
+++ b/src/singularity.ts
@@ -21,7 +21,7 @@ export const updateSingularityPenalties = (): void => {
                  Cube Upgrade Costs (Excluding Cookies) are multiplied by ${format(calculateSingularityDebuff('Cube Upgrades', singularityCount), 2, true)}.
                  ${platonic}
                  ${hepteract}
-                 Your penalties will sharply increase in <span style="color: red"> Singularity ${format(calculateNextSpike(player.singularityCount), 0, true)}</span>.
+                 Your penalties will ${singularityCount >= 250 ? 'now smoothly increase forever.' : `sharply increase in <span style="color: red"> Singularity ${format(calculateNextSpike(player.singularityCount), 0, true)}</span>.`}
                  <span style='color: ${color}'>Antiquities of Ant God is ${(player.runelevels[6] > 0) ? '' : 'NOT'} purchased. Penalties are ${(player.runelevels[6] > 0) ? '' : 'NOT'} dispelled!</span>`
 
     DOMCacheGetOrSet('singularityPenaltiesMultiline').innerHTML = str;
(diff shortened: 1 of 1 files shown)


src/singularity.ts
```diff
@@ -21,7 +21,8 @@ export const updateSingularityPenalties = (): void => {
                  Cube Upgrade Costs (Excluding Cookies) are multiplied by ${format(calculateSingularityDebuff('Cube Upgrades', singularityCount), 2, true)}.
                  ${platonic}
                  ${hepteract}
-                 <br><span style='color: ${color}'>Antiquities of Ant God is ${(player.runelevels[6] > 0) ? '' : 'NOT'} purchased. Penalties are ${(player.runelevels[6] > 0) ? '' : 'NOT'} dispelled!</span>`
+                 Your penalties will sharply increase in <span style="color: red"> Singularity ${format(calculateNextSpike(player.singularityCount), 0, true)}</span>.
```
mustache0110, line 24:
```suggestion
                 Your penalties will ${player.singularityCount >= 250 ? 'now smoothly increase forever.' : `sharply increase in <span style="color: red"> Singularity ${format(calculateNextSpike(player.singularityCount), 0, true)}</span>.`}
```

mustache0110: Just another way around the cap/end at 250, use it if you like it. Or add some dramatic flair. Crushing despair of Derpsmith or Ant God. All that.

src/singularity.ts
```diff
@@ -1471,7 +1472,32 @@ export const calculateEffectiveSingularities = (singularityCount: number = playe
...
+        PenaltyArray = 151
+    }
+    if (singularityCount > 150) {
+        PenaltyArray = 250
+    }
+    if (singularityCount === 250) {
+        PenaltyArray = 1337
+    }
 
+    return PenaltyArray
```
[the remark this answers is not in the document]
mustache0110: Again, use it if you like it. Getting around the cap/end at 250. You may want to return -1 outside the loop in case it gets called above array end. It never should, but code changes. And typescript will probably demand it, anyway.

Using the array format makes it easier(?) if more thresholds are added to just have to change fewer parts.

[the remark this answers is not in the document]
Mixelz: Added in this suggestion!

src/singularity.ts
Mixelz: Added this in!

src/singularity.ts
```diff
@@ -9,8 +9,8 @@ import { toOrdinal } from './Utility'
 export const updateSingularityPenalties = (): void => {
     const singularityCount = player.singularityCount;
     const color = player.runelevels[6] > 0 ? 'green' : 'red';
-    const platonic = (singularityCount > 36) ? `Platonic Upgrade costs are multiplied by ${format(calculateSingularityDebuff('Platonic Costs', singularityCount), 2, true)}.` : '';
-    const hepteract = (singularityCount > 50) ? `Hepteract Forge costs are multiplied by ${format(calculateSingularityDebuff('Hepteract Costs', singularityCount), 2, true)}.` : '';
+    const platonic = (singularityCount > 36) ? `Platonic Upgrade costs are multiplied by ${format(calculateSingularityDebuff('Platonic Costs', singularityCount), 2, true)}.` : '<span style="color: grey">???????? ??????? ????? ??? ?????????? ?? ???</span> <span style="color: red">(Sing 37)</span>';
+    const hepteract = (singularityCount > 50) ? `Hepteract Forge costs are multiplied by ${format(calculateSingularityDebuff('Hepteract Costs', singularityCount), 2, true)}.` : '<span style="color: grey">????????? ????? ????? ??? ?????????? ?? ???</span> <span style="color: red">(Sing 51)</span>';
```
KhafraDev, line 13: what's the purpose of having these elements?

KhafraDev (approved):

src/singularity.ts
Mixelz: As stated in the original commit, it's used to indicate at what singularity those two penalties unlock as shown in the example image I provided! Prior to this it was just being displayed as empty space if you did not have a high enough singularity for them to be activated, so I simply repurposed that for this. 
Sing-Example
I just went with the Question Marks as the baseline, we can change it to something like "A new penalty will unlock in Singularity X" if Plat wants. :)

PR #368, opened 2022-10-06
Octo must

mustache0110:
Separate commits in case you want to skip any... otherwise feel free to squash into single commit.

### DIFF
--- a/src/Synergism.ts
+++ b/src/Synergism.ts
@@ -870,6 +870,9 @@ const loadSynergy = async () => {
         Object.defineProperty(window, 'Decimal', {
             value: Decimal
         });
+        if (data && testing) {
+            data.exporttest = false;
+        }
     }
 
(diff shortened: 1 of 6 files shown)


src/Synergism.ts
```diff
@@ -870,6 +870,9 @@ const loadSynergy = async () => {
         Object.defineProperty(window, 'Decimal', {
             value: Decimal
         });
+        if (data && testing) {
+            data.exporttest = false;
+        }
```
mustache0110, line 875: Last time I pushed this with the intent of making it easier to test builds publicly. Looking at quark/GQ/Oct spending in particular and did not want saves to be ruined or bugs to award free quarks. Not that important in a single player game, but still did not want to expose easy warping and day resets to testers able to migrate back.

This version checks that we're actually testing, and not just non-production.

KhafraDev (approved):

PR #372, opened 2022-10-09
Fix Auto Catalyze instantly spending a lot of potions

dorijanko:
Fixes the most recent bug report in the discord server. If the autopotion timer is at, for example, 20/100 seconds and auto catalyze gets turned on, it will instantly spend 400 potions; this fixes that.


KhafraDev:
Thanks!

PR #374, opened 2022-10-11
Add open custom cubes until

Vonuras:
might need rewording for the buttons

### DIFF
--- a/src/CubeExperimental.ts
+++ b/src/CubeExperimental.ts
@@ -93,7 +93,8 @@ export abstract class Cube {
         const thisInPlayer = player[this.key] as Cube;
         const amount = await Prompt(
             `How many cubes would you like to open? You have ${format(thisInPlayer, 0, true)}! ` +
-            'You can input a percentage of cubes to open, for example: "50%" or "100%".'
+            '\nYou can input a percentage of cubes to open, for example: "50%" or "100%".' +
+            '\nYou can start the input with "-" (put in a negative value) to put in a value, that will be saved, which means you open all cubes except the specified number. (also works with %)'
         );
 
         if (amount === null) {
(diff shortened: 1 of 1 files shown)


core-exe:
wow thanks for implementing my suggestions!

mustache0110:
Do we need another button that does the same thing as Custom? I feel like you could just as easily put the functionality into Custom.
``` `How many cubes would you like to open? You have ${format(thisInPlayer, 0, true)}! ` +
            'You can input a percentage of cubes to open, for example: "50%" or "100%".' + '/n/n Prefix your number with an "S" to save the amount of cubes while opening the rest. "S12345" or "S1e12"  ```

Or use ~ or - as the prefix. I'm not sure how universally available ~ is on international keyboards, though. (S)ave, (R)eserve, (H)old are probably more accessible. Using the negative sign might be even better/easier. `Input a negative number to save/reserve/hold that amount while opening the rest.`

**Comment by Vonuras** [comment]
I force pushed the changes, to implement it the way you recommended.

Pretty nice idea to actually have it in the same alert window

Might still need rewording

**Comment by mustache0110** [comment]
I'm sorry, I didn't mean to suggest you include all three prefixes, I was just offering examples. I would say to just pick 1 to help cut down on the wording.

But, I like the negative number idea better now, the more I think about it. Because `toSpend = total + (-input)` is kind of cute and easy. Otherwise you have to parse more.

It might be as simple as just changing
```
const cubesToOpen = amount.startsWith('-')
    ? isPercentage ? 100 + Number(amount.slice(0, -1)) : thisInPlayer.value + Number(amount)
    : isPercentage ? Number(amount.slice(0, -1)) : Number(amount);
```
(Although I didn't test it out) Ultimately, it's not much different if you prefer a specific prefix.

Vonuras:
Changed it to your Idea with "-"
I actually really like that one as well

the text might need rewording (please change if someone finds a better wording)

PR #376, opened 2022-10-12
Added new Perk: Achievement Hoarder

jarboo:
Perk unlocked at Sing3. Description: "Some annoying to get achievements are kept on Singularity (like :unsmith: and :antismith:)"

I've chosen the 4 achievements that are annoying and sometimes don't unlock through normal play

image
image


jarboo:
@KhafraDev you don't like this PR? ;)

jahwsuf:
I don't mind those as much as I do the transcend + prestige ones.  Prestige in particular; by the time I remember, I'm usually under  too heavy a MD corruption to trigger 'em!

There's also the fact that awarding them early gives far less quarks than unlocking them with quark hept up.

PR #378, opened 2022-10-14
add all current mults to oct multiplier stats tab

Feodoric:
Previous oct multiplier tab was only showing 11 of the 20+ multipliers that get factored into octeracts. It also was using a different formula for the score multiplier than the real `OcteractGainPerSecond` function was using. 

I pulled the multiplier array up into its own method to avoid code duplication, added new DOM elements to the index.html file and expanded the octeract multiplier map in statistics.


KhafraDev:
thank you

PR #379, opened 2022-10-14
fix deleting save file if cancel on 0+0

Feodoric:
The prompt code returns `null` if the player clicks cancel, but that null was then being cast to the number 0. If the addition problem randomly rolled 0+0, this would result in a 'successful' solve of the addition confirmation. 

I added a null check on the prompt result to explicitly check for a cancel.


KhafraDev:
thank you

PR #380, opened 2022-10-14
divine pack now applies 7.77x instead of 2x

Feodoric:
Looks like the octeract gain per second method was using a value of 2x for the divine pack multiplier instead of the 7.77x in the pack description. The old octeract stats display was displaying the 7.77x value.

This PR replaces the line calculating a 2x multiplier with the one from the old display method that properly calculated 7.77. Not sure if we should prefer using the `getEffect().bonus` style of checking on upgrade status or if checking the level is ok.


Feodoric:
Fixed an issue raised by @Dorijanko in discord where my previous fix would award a 2x multiplier if average corruptions were >=14 even if the player didn't have divine pack.

Pseudonian:
The actual effect is correct- it should change the description instead

Feodoric:
@Pseudonian force-pushed an update after rebasing onto latest master.

* Reverted the 7.77x multiplier to 2x in calculate
* Modified divine pack description to the correct 2x
* Changed divine pack and Cx20 descriptions to clarify that the requirements are average corruption level at 14 (hopefully make 10/15* loadout after chonky corruptions easier to figure out for people trying to play blind)

PR #382, opened 2022-10-16
Singularity and Octeract shop QOL

mustache0110:
This is my solution to finding what is purchasable. It uses a temporary game-state variable G['shopEnhanceVision'] to recolor the shop upgrades as you mouseover and mouseout of the upgrade container. I used CSS filter effects, as to not require transparency on icons and/or have background colors make them look terrible and/or mess with the art. This also unintentionally made them colourblind-friendly.

If an upgrade is already at cap, it's sightly dimmed.
If an upgrade is unpurchasable due to cost or level requirement, it is greyed and slightly dimmed.
if it's purchasable, it's slightly brightened and inverted.
If it's soft-capped, it's slightly brightened, inverted, blurred and saturated.

https://mustache0110.github.io/SynergismOfficial/ to see it in action. It helps remove the need for a hide max, and can easily add a opacity(0) to the first condition to hide them. But this is an afterthought, and does not honor the art.


KhafraDev (approved):

mustache0110:
image
Mouseout state

image
Mouseover state

PR #383, opened 2022-10-19
October fix

mustache0110:
minor cleanup, no real logic changes. stripped the comments out because the functions are well named enough to not be necessary, (more importantly they staggered the actual code)

Daily code Confirm/Alert bug reported by randomtuba


src/ImportExport.ts
```diff
@@ -677,7 +677,7 @@ const addCodeInterval = () : number => {
     let time = hour
     time *= (1 - 0.02 * player.shopUpgrades.calculator4)
     time *= (1 - Math.min(.5, (player.highestSingularityCount >= 125 ? player.highestSingularityCount / 1000 : 0)
-                            - (player.highestSingularityCount >= 200 ? player.highestSingularityCount / 1000 : 0)))
+                            + (player.highestSingularityCount >= 200 ? player.highestSingularityCount / 1000 : 0)))
```
mustache0110, line 680: Bug introduced when I neglected to change the sign when wrapping it in the Math.min() call.

PR #384, opened 2022-10-20
Octeract upgrade: added time to afford next level

jarboo:
When displaying an Octeract upgrade details, we now show if the next level if affordable, or  how long it will take to afford it, using current Oct/sec production
image
image
image

### DIFF
--- a/src/Octeracts.ts
+++ b/src/Octeracts.ts
@@ -1,9 +1,10 @@
-import { format, player } from './Synergism';
+import { format, player, formatTimeShort } from './Synergism';
 import { Alert, Prompt } from './UpdateHTML';
 import type { IUpgradeData } from './DynamicUpgrade';
 import { DynamicUpgrade } from './DynamicUpgrade';
 import type { Player } from './types/Synergism';
 import { DOMCacheGetOrSet } from './Cache/DOM';
+import { octeractGainPerSecond } from './Calculate'
 
 export interface IOcteractData extends IUpgradeData {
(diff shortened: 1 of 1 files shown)


src/Octeracts.ts
```diff
@@ -107,11 +109,21 @@ export class OcteractUpgrade extends DynamicUpgrade {
             freeLevelInfo = freeLevelInfo + '<span style="color: maroon"> (Softcapped) </span>'
         }
 
+        const isAffordable = costNextLevel <= player.wowOcteracts;
+        let affordTime = '';
+        if (!isMaxLevel && !isAffordable) {
+            const octPerSecond = octeractGainPerSecond();
```
KhafraDev, line 115: when could this be less than or equal to 0?

KhafraDev (changes requested):

src/Octeracts.ts
jarboo: When ascension score is too low (typically at the start of each Sing), Octeract production is 0.
I reused the 'Infinity' term that is used to show Oct production

KhafraDev: https://github.com/Pseudo-Corp/SynergismOfficial/blob/72b56eefad93e099ace295cc1d7a10bb2366cd11/src/Calculate.ts#L1438

1 / anything can never equal 0 in js, or am I mistaken?

KhafraDev (approved):

PR #385, opened 2022-10-20
Change number of digits displayed by format function

jarboo:
The format() function will now switch to the "mantissa e power" notation sooner: 1e7 instead of 1e13. This change will make visually comparing big numbers easier. This is a global change.

Octeract page before:
image
After:
image

Singularity page before:
image
After:
image


KhafraDev (approved):
@Pseudonian

PR #386, opened 2022-10-24
Fancy fraction formatting

mustache0110:
2 reasons behind this feature:
1. A number of people expressed confusion with octeract costs because they're used to seeing the / as a level separator.
2. I like how it looks
image

I dropped in a flexible helper function, but it could be simplified. If @Pseudonian wants to put it directly in format, we can just have the fractional code output .innerHTML code. Then for the few elements that use fractional code, calling DOM('element').innerHTML = format(sometimes fractional number) will work fine when the number is big or small. I'll push this idea as a separate commit, because simpler might be better.

The more complicated change could avoid innerHTML use by showing/hiding a fractional container and setting the numerator and denominator as .textContent. I don't think this is necessary, I'm not sure.


mustache0110:
- Fixed merge conflict with recent format( ) changes.
- Adjusted font size for better readability; changed text alignment to be much less disruptive to the line above.

Font size could be nudged up or down, but current size seems to be a good balance between standing out and legibility.

PR #387, opened 2022-10-27
DIsplay C15 exponent instead of C15 completions

jarboo:
DIsplay C15 exponent instead of C15 completions
image

When C15 hasn't been run in the current Sing, it looks like this:
image

### DIFF
--- a/src/UpdateHTML.ts
+++ b/src/UpdateHTML.ts
@@ -861,7 +861,11 @@ export const updateChallengeLevel = (k: number) => {
     const el = DOMCacheGetOrSet('challenge' + k + 'level');
     const maxChallenges = getMaxChallenges(k);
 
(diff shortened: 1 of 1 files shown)


src/UpdateHTML.ts
```diff
@@ -861,7 +861,11 @@ export const updateChallengeLevel = (k: number) => {
     const el = DOMCacheGetOrSet('challenge' + k + 'level');
     const maxChallenges = getMaxChallenges(k);
 
-    el.textContent = `${player.challengecompletions[k]}/${maxChallenges}`;
+    if (k === 15) {
+        el.textContent = format(player.challenge15Exponent,0,true);
```
KhafraDev, line 865:
```suggestion
        el.textContent = format(player.challenge15Exponent, 0, true);
```

PR #389, opened 2022-10-30
fixed PL-AT Σ

httpsnet:
PL-AT Σ becomes level 2 at 200th Singularity, but loses its effectiveness.
Then the explanation of the displayed text is different, so I think it's a bug.
PL-AT Σ - Level 2 may have intended its effect to be +100%, so PR will now change the calculation from - to + and update the description.

### DIFF
--- a/src/singularity.ts
+++ b/src/singularity.ts
@@ -1264,8 +1264,15 @@ export const singularityPerks: SingularityPerk[] = [
     {
         name: 'PL-AT Σ',
         levels: [125, 200],
-        description: () => {
-            return 'Code \'add\' refills 0.1% faster per level per singularity (MAX: 50% faster)'
+        description: (n: number, levels: number[]) => {
+            let counter = 0
+            for (const singCount of levels) {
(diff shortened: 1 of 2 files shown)


src/singularity.ts
```diff
@@ -1264,8 +1264,15 @@ export const singularityPerks: SingularityPerk[] = [
     {
         name: 'PL-AT Σ',
         levels: [125, 200],
-        description: () => {
-            return 'Code \'add\' refills 0.1% faster per level per singularity (MAX: 50% faster)'
+        description: (n: number, levels: number[]) => {
+            let counter = 0
+            for (const singCount of levels) {
+                if (n >= singCount) {
+                    counter += 1
```
KhafraDev, line 1271: you could add 0.1 to counter and remove the / 10

httpsnet: fixed

httpsnet:
The calculation was corrected by the October fix. Currently only changing the message

PR #390, opened 2022-10-31
quick reversion

mustache0110:
I missed some of the effects of this change, they did not show up with the tests i ran. And while it fixed some problems, it causes more. I'm pretty sure I can still fix the bug while not causing more, but it will required more time for testing/rewriting. While i'm working on that, this will at least revert to the old behavior.


KhafraDev:
what bug(s) did this cause?

mustache0110:
C6-C9 reset for sure

KhafraDev:
can you fix the linting issues?

PR #392, opened 2022-11-02
Fix Singularity

httpsnet:
Fixes that current Singularity has an issue that can go up to s254, allowing you to avoid Singularity 250 penalty.
Fix not being able to Singularity after achieving Singularity 250.

Singularity 250 penalty changes from 250 to 250+.
If the next Singularity exceeds 250, make it 250.
If your current Singularity is 250 or higher, continue to increase your Singularity count.


httpsnet:
Conflict fix
Again capped at Sing 250
Stop at the estimated Sing count even with Fast Forward. set to s200 and s250

PR #393, opened 2022-11-02
Revert overly zealous shop confirmations

pingubass:
#340 made it so that when you buy refundable shop items and hit max level, there is an alert, even if you have shop confirmations off. This means that doing a shop respec cycle between cubes and speedups and off/obt/ant, it is no longer possible to just hold the enter key and click around the shop screen because you have to dismiss an alert each time it fires.

This PR is a suggestion to fix it, however I have not tested it - I am not sure how to set up a test environment, so would love someone else to test it if possible? @mustache0110 maybe?

### DIFF
--- a/src/Shop.ts
+++ b/src/Shop.ts
@@ -806,11 +806,20 @@ export const friendlyShopName = (input: ShopUpgradeNames) => {
 
 export const buyShopUpgrades = async (input: ShopUpgradeNames) => {
     const shopItem = shopData[input];
-    if (player.shopUpgrades[input] >= shopItem.maxLevel) {
-        return Alert(`You can't purchase ${friendlyShopName(input)} because you are already at the maximum ${shopItem.type === shopUpgradeTypes.UPGRADE ? 'level' : 'capacity'}!`);
-    } else if (Number(player.worlds) < getShopCosts(input)) {
-        return Alert(`You can't purchase ${friendlyShopName(input)} because you don't have enough Quarks!`);
+
+    const maxLevel = player.shopUpgrades[input] >= shopItem.maxLevel;
+    const canAfford = Number(player.worlds) >= getShopCosts(input);
+    if (maxLevel || !canAfford) {
+        if (player.shopConfirmationToggle || (!shopItem.refundable && player.shopBuyMaxToggle)) {
+            if (maxLevel) {
+                return Alert(`You can't purchase ${friendlyShopName(input)} because you are already at the maximum ${shopItem.type === shopUpgradeTypes.UPGRADE ? 'level' : 'capacity'}!`);
+            } else if (!canAfford) {
+                return Alert(`You can't purchase ${friendlyShopName(input)} because you don't have enough Quarks!`);
+            }
+        }
+        return;
     }
+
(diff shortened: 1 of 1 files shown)


mustache0110:
I wouldn't even go that far; I see those alerts as new player warnings. And we do want to protect new players from wasting quarks that they can't refund. So while I do wish shopconfirmation toggle could eliminate more alerts, we also do want to protect the new players. I hadn't been able to come up with a good solution, but you gave me a good idea.

Those two alerts could easily be just eliminated on shopconfirmation toggle because they do not waste quarks. I missed that check when I made those return early. My suggested change is below, which I can't display as code and also highlight the changes. The two return lines check the toggle and return an alert or null.

```js
    if (player.shopUpgrades[input] >= shopItem.maxLevel) {
        return player.shopConfirmationToggle ? Alert(`You can't purchase ${friendlyShopName(input)} because you 
are already at the maximum ${shopItem.type === shopUpgradeTypes.UPGRADE ? 'level : null;
' : 'capacity'}!`)
    } else if (Number(player.worlds) < getShopCosts(input)) {
        return player.shopConfirmationToggle ? Alert(`You can't purchase ${friendlyShopName(input)} because you 
don't have enough Quarks!`) : null;
    }
```

src/Shop.ts
```diff
@@ -806,11 +806,20 @@ export const friendlyShopName = (input: ShopUpgradeNames) => {
...
+    const canAfford = Number(player.worlds) >= getShopCosts(input);
+    if (maxLevel || !canAfford) {
+        if (player.shopConfirmationToggle || (!shopItem.refundable && player.shopBuyMaxToggle)) {
+            if (maxLevel) {
+                return Alert(`You can't purchase ${friendlyShopName(input)} because you are already at the maximum ${shopItem.type === shopUpgradeTypes.UPGRADE ? 'level' : 'capacity'}!`);
+            } else if (!canAfford) {
+                return Alert(`You can't purchase ${friendlyShopName(input)} because you don't have enough Quarks!`);
+            }
+        }
+        return;
```
mustache0110, line 820:
```suggestion
    if (player.shopUpgrades[input] >= shopItem.maxLevel) {
        return player.shopConfirmationToggle ?
        Alert(`You can't purchase ${friendlyShopName(input)} because you are already at the maximum ${shopItem.type === shopUpgradeTypes.UPGRADE ? 'level' : 'capacity'}!`) :
        null;
    } else if (Number(player.worlds) < getShopCosts(input)) {
        return player.shopConfirmationToggle ? Alert(`You can't purchase ${friendlyShopName(input)} because you don't have enough Quarks!`) : null;
```
If you want to slightly tame that super long line of code, I believe this is the style Plat has used in the past.

pingubass:
Your suggestion seems totally sane to me, I was just reverting the behaviour to what it was before. I guess the `!shopItem.refundable && player.shopBuyMaxToggle` part says to show the alert anyway on non-refundable items if the MAX toggle was on. As you say, it won't waste quarks, so is not necessary. Let me fettle the code...

KhafraDev (approved):

PR #394, opened 2022-11-03
better check on 'ONE MIND' upgrade name

mustache0110:
also changed two Alerts to not return, so that the tooltips update/reveal code below also runs.


mustache0110:
Tested on both a fresh save with [NEW!] 'ONE MIND', and a refreshed save with [56] 'ONE MIND' . Both easily verified to work due to the Alert popping up.

Pseudonian (approved):

PR #398, opened 2022-11-08
fix formats in new notations

Vonuras:


src/Synergism.ts
```diff
@@ -1877,7 +1877,14 @@ const locOpts = { minimumFractionDigits: 2, maximumFractionDigits: 2 };
 
 const padEvery = (str: string, places = 3) => {
     let step = 1, newStr = '';
-    for (let i = str.length - 1; i >= 0; i--) {
+    let indexOfPoint = str.indexOf('.');
+    if (indexOfPoint == -1) {
```
KhafraDev, line 1881:
```suggestion
    if (indexOfPoint === -1) {
```

src/Synergism.ts
```diff
@@ -1887,6 +1894,10 @@ const padEvery = (str: string, places = 3) => {
         }
     }
 
+    // re-add decimal places
+    if (indexOfPoint != str.length - 1) {
```
KhafraDev, line 1898:
```suggestion
    if (indexOfPoint !== str.length - 1) {
```

KhafraDev (changes requested):

Vonuras:
reverted the undefined check, as linter fails otherwise

PR #404, opened 2022-11-14
Theme's rework (cleaned up)

awWhy:
Everything is same


awWhy:
Don't merge yet (if you are reading), I need to change some colors like orchid to var(--orhchid-text-color), that came with challenges

awWhy:
There is now nothing that can affect game, minus that theme's were moved into local strorage instead of save file

Pseudonian (approved):

PR #405, opened 2022-11-14
Daily code stats

Feodoric:
* Added a new utility function that can handle reducing arrays that have both additive and multiplicative bonuses.
 * Moved the calculation for daily code free upgrade rolls out of ImportExport.ts and into Calculate.ts. The new function returns a `{list, total}` object like the similar functions for cube multipliers etc.
 * Added a new stats for nerds block to display the daily code free upgrade roll bonuses and total.


Feodoric:
This new stats block has been requested a few times in the #s100-haha-bbshard channel of the discord, figured I'd give it a shot.

PR #410, opened 2022-11-21
Fixed generated Dots not being reset on singularity

Yayformee:


mustache0110:
Definitely a typo / bug.

I'm not sure what iteration of the code this is from, probably pre-actual-Singularity content. Rather than reset the many things that need to be reset, the game now loads a blank save to start a singularity, and copies over the things it wants to keep. Currently, the buildings are reset due to the blank save, although hard to tell since they're hidden ;d

So, this code is like a vestigial tail. Good eye, though. It's still a fix, even if the code is dead. I would merge it, even if it's unnecessary, because it is technically correct

Yayformee:
I only noticed it while reviewing another change that I was making.

Would it be alright for me to go and remove the dead code then? I will make a separate pull request.

PR #412, opened 2022-11-23
Event: tiny change

mustache0110:
99% sure this is the intended behavior

almost not worth a PR, but I figured this is more noticeable than a PM in the sea of discord pings


mustache0110:
@Pseudonian

mustache0110:
empty commit

PR #417, opened 2022-12-08
Add automatic upgrades

httpsnet:
This PR adds Automation Cube Upgrades and Automation Platonic Upgrades.
When toggled on, upgrades will be made for you if you have the resources to pay for upgrades you didn't purchase.
The order in which upgrades are performed is in order of simple cost.
Automation Platonic Upgrades do upgrades without the cost of Offerings and Obtainiums.
Upgrade: 1 Level wow in Automation Cube Upgrades only buy if you have enough cubes to upgrade to max level at once.
Each of these works with Ascension.

Automation Cube Upgrades - Level 1 requires Singularity 50.
Automation Cube Upgrades - Level 2 requires Singularity 150.
Automation Platonic Upgrades - Level 1 requires Singularity 100.
Automation Platonic Upgrades - Level 2 requires Singularity 200.

Level 1 works only if the current Singularity is less than the Highest Singularity.
So it will work in Singularity Challenge.
Level 2 overrides Level 1 functional restrictions. In other words, it will work even with Highest Singularity.
Both automations are heavily rewarded because they greatly improve the QoL of Singularity.

other fixes
Fixed an issue where if player.toggles[28] is True, it was possible to Ascension even if Ascension was not unlocked by using the Ascension hidden button or the Ascension hotkey. Also toggles has been fixed to 31.
Can no longer work if Autocraft and Auto open Cubes unlock conditions are not met.
Removed Research 183 from Auto open Cubes Hypercubes requirement. This fixes a non-functional issue early in C13 to C15.


awWhy:
Just in case remind that need to remove "&#013Research 8x8 is required" from html line 2265

Pseudonian (approved):

PR #420, opened 2022-12-10
Fix css and swap img element into input

awWhy:
Just removing width: auto, and setting it to fixed value fixes all issues (to make it responsive just change px to vw)

Showcase
After
image

Before
image


awWhy:
Added HTML fix, because I want to be able to use keyboard on images
Changed img element into input, removed in these places any table's, fix css, also removed labels

awWhy:
I'm going to reopen it after spliting css\html and SR, into 2 different PR's (because Plat will more likely accept fix to html this way)

PR #426, opened 2022-12-24
feature: adds an add-code stat page

jahwsuf:
The picture of a thousand words:

image

Introducing a new "Stats for Nerds" page - "Add Code Stats"!

I selected the 'green' color for the button due to its color under "Special Actions" on the Settings tab.  No reason it can't be tweaked if desired, of course.

I intentionally left out the Midas perk; it'd clutter the page somewhat and it's not currently modified by any other in-game effects, so there's little need to show its details here.

Ideally, the "bonus octeract time" should probably be "bonus octeracts" - like, the amount that would be earned.  I don't have the calculator unlocked in my save in order to test that properly, though.

----

In order to gain access to the various effects that are enumerated on the page, I refactored existing add-code related methods to resemble many of the methods in calculate.ts that are used for both in-game calculations and stats reports.  I thought about maybe moving them to calculate.ts, but felt that wasn't the right call for this PR, as it'd make code review a little harder to perform.

----

The primary reason I decided to work on this?  The recent change to add code duration in higher-level singularities, which made it much harder to infer these values once related singularity perks kick in.

This is also why I chose to have the page unlock once the Singularity tier is reached.  Sure, it could be set to unlock earlier, but the complicating effects aren't added until singularity.


index.html
```diff
@@ -3330,6 +3331,48 @@ <h1>Artists</h1>
                     id="sOcMT" alt="sOcMT" label="sOcMT" class="statNumber statTotal">0</span></p>
         </div>
 
+        <div id="addCodeStats" alt="addCodeStats" label="addCodeStats" class="statContainer" style="display: none;">
+            <p id="addCodeStatTitle1" alt="addCodeStatTitle1" label="addCodeStatTitle1" class="statPortion" style="color: green; font-size: 1.2em">Add Code Effects</p>
+            <p id="stat+eff1" alt="stat+eff1" label="stat+eff1" class="statPortion singularity">Quarks:<span id="s+eff1" alt="s+eff1" label="s+eff1" class="statNumber"></span></p>
```
jahwsuf, line 3336: I think I need to remove the `singularity` bit on most of these; the reason for that part of the classlist _just_ clicked for me.

While working on a second stats page for global speed, since that's been requested a bit.

KhafraDev (approved):

PR #432, opened 2022-12-31
Added Color to Sing Penalties text

Mixelz:
Added some simple color to the penalty block, this is meant to help with readability and make the tab at more aesthetically pleasing. Chosen colors should also have no issues with colorblind folks and mostly match established color patterns already (Ascension Speed is Orange, Hepteract Forge is Pink, etc.) 
image


awWhy:
Not very important, but in order to make orchid text to be readable and changeable by theme's
Need to change 'style: Orchid' into 'style: var(--orchid-text-color)'

PR #433, opened 2023-01-01
change: restyles sing shop + oct shop enhanced vision

jahwsuf:
The styling introduced for Singularity + Octeract shop "enhanced vision" was met with mixed response in regard to its styling.  The change on hover is quite extreme and takes a bit to get used to, after all.

If nothing else, I hope this prompts some design discussion on what better styling might be & how to make these things to the user without requiring "jarring" styling effects.

------

This proposal keeps the core styling about the same, opting for tweaks that are generally more consistent with other in-game styles where possible.  Where not possible, it opts for commonly-used visual cues if possible.

So, how would this proposed alternative style change things?

image

After some fun CSS experimentation, I figured out a way to give a faux green / "completed" background to fully-purchased shop items.  The trick:  the image element is given partial opacity, while its _parent_ element is given a green background. 
- Note that the green background can't be put on the image element.

The faded crimson background replaces the "blur" styling; that crimson is the same crimson used for the "softcapped" text (as seen in the image) after all filter effects are applied.  Said other effects aim to make the options "pop" more while also reducing opacity in order to composite the 'filtered' version with the original.  (Same image on two layers, one altered, one not - then blended together.)

Items at approximately normal coloration and brightness - they're fully purchasable.  At "half" coloration and brightness, they're "in reach", but the player doesn't have enough quarks.

Finally, the dulled-out, desaturated options are simply unreachable at the current max singularity count ever reached by the player.

----

Oh, and for octeracts:

image

Very similar styling has been applied, though with one notable tweak.  Since octeract bonuses have no minimum singularity count, what to do?

"Half" coloration / brightness is used for perks that are, or have been, attainable to purchase at some point - even if only on previously-attained levels.  Note that the highlighted perk, Vitamin O, has purchased levels.  Here in particular, it costs less octeracts than the player's lifetime total as well.  So, on two counts, this particular oct perk isn't hard for the player to attain.

The dulled-out, desaturated options are considered fully unreachable by the player.  The rule of thumb I used?  If the base level of the perk requires more octeracts than the player has ever generated, it's completely out of reach.

----

It's probably worth note that I am currently unable to test the "completed" styling against the perks with a strong pre-existing background color - some testing may be needed for those.

I've put this PR in draft for now mostly because it could probably use a community review or something, rather than triggering a chain of tweaks across multiple releases.  I don't mind undrafting it if desired, though.

### DIFF
--- a/src/UpdateVisuals.ts
+++ b/src/UpdateVisuals.ts
@@ -560,16 +560,28 @@ export const visualUpdateSingularity = () => {
                 continue
             }
             const singItem = player.singularityUpgrades[key];
-            const el = DOMCacheGetOrSet(`${String(key)}`);
+            const el = DOMCacheGetOrSet(`${String(key)}`) as HTMLImageElement;
             if (singItem.maxLevel !== -1 && singItem.level >= singItem.computeMaxLevel()) {
-                el.style.filter = val ? 'brightness(.9)' : 'none';
-            } else if  (singItem.getCostTNL() > player.goldenQuarks || player.singularityCount < singItem.minimumSingularity) {
-                el.style.filter = val ? 'grayscale(.9) brightness(.8)' : 'none';
+                // Upgrade is maxed; provide a makeshift 'completed green' look for these entries.
+                el.style.filter = val ? 'opacity(0.6)' : 'none';
+                el.parentElement!.style.backgroundColor = val ? 'green' : '';
+            } else if (player.singularityCount < singItem.minimumSingularity) {
+                // Upgrade is absolutely unpurchasable
+                el.style.filter = val ? 'saturate(.15) brightness(.6)' : 'none';
+            } else if (singItem.getCostTNL() > player.goldenQuarks) {
+                // Upgrade is not currently purchasable.
+                el.style.filter = val ? 'saturate(.6) brightness(.8)' : 'none';
             } else if (singItem.maxLevel === -1 || singItem.level < singItem.computeMaxLevel()) {
                 if (singItem.freeLevels > singItem.level) {
-                    el.style.filter = val ? 'blur(1px) invert(.9) saturate(200)' : 'none';
+                    // Give it both a hint of enhanced saturation/brightness and a translucent drop-shadow
+                    // using the standard color for "softcapped" text.
+                    el.style.filter = val ? 'saturate(1.6) brightness(2) opacity(0.33) drop-shadow(0 0 1px var(--crimson-text-color))' : 'none';
+                    // Because we made the base image transparent, we replicate it in its parent's background to keep the image itself
+                    // at full opacity when composited.
+                    el.parentElement!.style.backgroundImage = val ? `url(${el.src})` : '';
                 } else {
-                    el.style.filter = val ? 'invert(.9) brightness(1.1)' : 'none';
+                    // Eh, just give it a subtle bit of extra pop, no other tweaks needed.
+                    el.style.filter = val ? 'saturate(1.2) brightness(1.1)' : 'none';
                 }
             }
         }
@@ -580,16 +592,28 @@ export const visualUpdateSingularity = () => {
 
         for (const key of keys) {
             const octItem = player.octeractUpgrades[key];
-            const el = DOMCacheGetOrSet(`${String(key)}`);
+            const el = DOMCacheGetOrSet(`${String(key)}`) as HTMLImageElement;
             if (octItem.maxLevel !== -1 && octItem.level >= octItem.maxLevel) {
-                el.style.filter = val ? 'brightness(.9)' : 'none';
-            } else if  (octItem.getCostTNL() > player.wowOcteracts) {
-                el.style.filter = val ? 'grayscale(.9) brightness(.8)' : 'none';
+                // Upgrade is maxed; provide a makeshift 'completed green' look for these entries.
+                el.style.filter = val ? 'opacity(0.6)' : 'none';
+                el.parentElement!.style.backgroundColor = val ? 'green' : '';
+            } else if (octItem.level == 0 && octItem.getCostTNL() > player.totalWowOcteracts) {
+                // Upgrade is not currently purchasable and has never been purchasable, even with 100% investment.
+                el.style.filter = val ? 'saturate(.15) brightness(.6)' : 'none';
+            } else if (octItem.getCostTNL() > player.wowOcteracts) {
+                // Upgrade is not currently purchasable, but might be reasonable to wait for.
+                el.style.filter = val ? 'saturate(.6) brightness(.8)' : 'none';
             } else if (octItem.maxLevel === -1 || octItem.level < octItem.maxLevel) {
+                // Give it both a hint of enhanced saturation/brightness and a translucent drop-shadow
+                // using the standard color for "softcapped" text.
                 if (octItem.freeLevels > octItem.level) {
-                    el.style.filter = val ? 'blur(2px) invert(.9) saturate(200)' : 'none';
+                    el.style.filter = val ? 'saturate(1.6) brightness(2) opacity(0.33) drop-shadow(0 0 1px var(--crimson-text-color))' : 'none';
+                    // Because we made the base image transparent, we replicate it in its parent's background to keep the image itself
+                    // at full opacity when composited.
+                    el.parentElement!.style.backgroundImage = val ? `url(${el.src})` : '';
                 } else {
-                    el.style.filter = val ? 'invert(.9) brightness(1.1)' : 'none';
+                    // Eh, just give it a subtle bit of extra pop, no other tweaks needed.
+                    el.style.filter = val ? 'saturate(1.2) brightness(1.1)' : 'none';
                 }
             }
         }


mustache0110:
Trying both side-by-side, I would say that the upgrades that you can buy are the most important information you want on mouse-over. They need to be the thing that pops most, that draws the most attention. A green background on "already purchased upgrades that you can't buy more of", i.e. something you don't care about, shouldn't be drawing more attention by being the thing that visually changes most. Crimson/drop shadow are hard to notice. Honestly, I had wanted to do more too but it was hard to make effects different enough to be noticeable without being too annoying. We need a \<blink\> tag for images. That's good UX!

Image you're old, or colorblind. Blur your eyes as you look at the screen, or look at the monitor from twice as far away. Invert wasn't what I wanted at first either. I tried many iterations, and it was the best **that I was able to come up with**. You want the button to say, 'hey click on me', and I believe it does. Part of the reason I liked my version was that it was pretty easy to refine, and I'm very much open to improvement.

It's a UI/UX problem. Why are you mousing over the grid? To buy something. So you want to draw attention to those exact items. I don't know what style Platonic prefers. If I did, I would just implement it because it's his game and he's in charge. It's obviously all subjective and you might prefer your version while I prefer mine at the same time. However, I firmly believe that any change you make should draw the most attention to the purchasable upgrades. Something like adding a Golden (GQ) border around said upgrades would also work if you really want to keep the green.

jahwsuf:
> It's a UI/UX problem. Why are you mousing over the grid? To buy something. So you want to draw attention to those exact items.

I agree with this, but I'd like to add one extra aspect worth consideration:  I'd like to recognize the items I was planning to buy when I mouse over them.

The visual change and contrast are so extreme I generally have to back off, double-check the item and location, and then click based on the location.  While it's nice to visually filter out all the clearly-irrelevant upgrades, I find I no longer recognize the ones that actually are relevant - in particular, the _most_ relevant given the circumstance.

Here's a screenshot I just took with mouse-over highlighting from the live version:

image

How easily can you identify the upgrades at row 2, column 5 and at row 3, column 2?  To me, they're far too distorted compared to their base appearance... and the blur filter that's part of the styling doesn't exactly help me recognize the original shapes involved.

----

> A green background on "already purchased upgrades that you can't buy more of", i.e. something you don't care about, shouldn't be drawing more attention by being the thing that visually changes most.

This is likely subjective, but I could argue that while that may happen for a split-second, afterward the green-background items will not be the most visually distinct, and thus will not draw more attention than the upgrades which maintain their brightness and saturation.

To be extremely explicit about my underlying assumption:  Things that are more visually distinct draw more attention than those that are less visually distinct.

For a side-by-side comparison:

**With mouse-over**:

image

**WIthout mouse-over**:

image

Note how the green-background items are faded in comparison to their non-hovered versions.  Fading something into black nigh-inherently makes it less visually distinct.

The most clear and prominent upgrade icons are those that are purchasable.  I realize that the softcapped items may be slightly less distinct than the non-softcapped ones; that is a hard balance to strike.

So, returning to that prior quote:

> It's a UI/UX problem. Why are you mousing over the grid? To buy something. So you want to draw attention to those exact items.

My proposition:  rather than _actively_ draw attention to the exact items, the goal is to reduce the screen's visual noise and provide clear visual "filters" that allow the player to far more easily focus their attention on those items.  In other words... to cause _everything else_ to "step aside" and stop demanding attention.

In _other_ other words:

> So you want to let the player's attention easily focus on, and easily recognize, those exact items.

In regard to all the above notes, for the full changeset of this PR and what I believe to be the approximate order of visual distinctiveness:

1.  Upgrades with their full/standard saturation and brightness levels (softcapped and non-softcapped)
2.  Upgrades that have been fully purchased.
3.  Upgrades that are "nearly purchaseable" (octs) / unlocked (singularities)
    - See Cookies 4 (row 2, col 3), the oct "sell out" upgrade (row 3, col 10), the potion upgrade to its right for examples in the comparison above.
4.  Upgrades that are effectively unpurchasable.

I suppose it is possible that items 1 and 2 are inverted.  That said, if so, I'd like to point to the upgrades, achievements, research, and cube upgrade tabs for the reasons I don't expect the green to distract players for long.  That is, by far, the most standardized visual cue when it comes to upgrades. I've long since learned to "filter out" green upgrades when I'm looking for what to purchase... or to _focus_ on the green upgrades when I want to review my progress toward current objectives.

----

> Image you're old, or **colorblind**.

As for the changes currently in this PR, I'll admit that I didn't think about red vs green colorblindness - that probably would matter with the backdrop and is a good point; a better color should be selected there.

- [ ] Better color selection for softcapped items re: colorblindness

mustache0110:
> Generally speaking, there may be multiple purchasable upgrades, while I'm currently planning to purchase just one type. I'm roughly familiar enough with the upgrades list to remember about where the item I want to purchase is, and thus I usually have visually "locked onto it" before mousing over. The issue: at that point, the current in-game mouse-over formatting completely changes its appearance and actively draws my attention to _all_ of the purchasable upgrades, breaking my visual lock-on. It's as if the styling used "flares", to use an airplane-combat metaphor. (Just in case, to be explicit, the styling actively and immediately diverts my attention away from what it had been on to focus on something - well, something**s** else. The original item _is_ included, yes, but that split-second is easily enough to break the lock-on.)

If you're at the point where you have all the shop items memorized and know what you want to buy, I think you can manage clicking on a square that does not move. I agree that the saturation effect and blur effect does distort the icons, but that was the level they needed to be to be visually distinct enough (IMO). My design angle was that the information is in the tooltip box below, and the icon change is to guide the mouse cursor and reveal that tooltip. Honestly, I'm not even sure softcapping means anything anymore because they run away faster than you can catch them anyway. Could easily just remove the blur/saturate, if that's the main sticking point. Originally, my thought was they give you more bang for your GQ, and deserved special attention, but the cost discrepancies, Sing level unlocks, and runaway softcap levels make the extra bang not so significant.

My bias is also to the new player, to people new to the mechanics, and the most frequent questions in discord channels. Not necessarily relevant here, but either I notice a rough edge personally, or the frequency of the same question / irk draws me to it.  You plop down 61 Singularity upgrades that are not sorted by cost, are also level locked, and it's hard to tell when you can buy and when. Researches also suffer from this; but it's a more farmable resource, and then the autobuyer steps in. Wowcube upgrades also have a similar issue, but again it's another relatively quickly farmable resource. GQ are a little more precious.

You shouldn't have to be on discord to play the game, the game should be able to help guide you. In general it does, but the Sing Shop felt a little lacking. Hence the (old) PR.
> > It's a UI/UX problem. Why are you mousing over the grid? To buy something. So you want to draw attention to those exact items.
> 
> I agree with this, but I'd like to add one extra aspect worth consideration: I'd like to recognize the items I was planning to buy when I mouse over them.
> 
> Compounding the noted issue above, the visual change and contrast are thus so extreme that I generally find I have to back off, double-check the item and location, and then click based on the location, actively ignoring the formatting so that my lock-on doesn't break again. While it **is** nice to visually filter out all the clearly-irrelevant upgrades (and you'll note I generally preserved that styling) I generally find that I no longer recognize the ones that actually are relevant - in particular, the _most_ relevant - given the circumstance.

Feels like a straw-man. If you know what the upgrades do based on the icon, you're already far enough along to know what they do base on their spot in the gird. Inversion doesn't change the art, and should be recognizable. Blur and saturate certainly do distort, but nothing refows the grid. Originally I had considered incorporating a hide max, but changing the grid was too disorienting.
> 
> Here's a screenshot I just took with mouse-over highlighting from the live version:
> How easily can you identify the upgrades at row 2, column 5 and at row 3, column 2 without referencing a different screen or image without the mouseover formatting?

Of course ;d But to mouseout is easy, or mouseover the actual icon gives you the description in the infobox below. I feel like you either know the upgrades or you don't. If you know them, it clearly doesn't matter. If you don't know them, then you have to mouseover every single one. Cutting down the relevant mouseovers from 61(+) to 5-10 is a good thing. 
> For me... not easily at all. They're far too distorted compared to their base appearance... and the blur filter that's part of the styling doesn't exactly help me recognize the original shapes involved.

Yes blur/saturate is weird and probably not necessary. Although I still kinda like it for the blueness. Art dosen't have to always be practical.
> > A green background on "already purchased upgrades that you can't buy more of", i.e. something you don't care about, shouldn't be drawing more attention by being the thing that visually changes most.
> 
> This is likely subjective, but I could argue that while that may happen for a split-second, afterward the green-background items will not be the most visually distinct, and thus will not draw more attention than the upgrades which maintain their brightness and saturation.

I don't think I expressed this point understandably. Which is sad because it was my main and only point. When I said that I did a side by side comparison, what I did was mouse in and out repeatedly. And when you do that, the green squares absolutely draw more attention. The purchasable upgrades are the only things that don't change, and that's the opposite of what you want to communicate. I maintain that you need to mouseover the individual icon to see cost/effect/buy anyway, and the highlighting is a guide to help achieve that.
> To be extremely explicit about my underlying assumption: Things that are more visually distinct draw more attention than those that are less visually distinct.

I agree, this is why invert (which is probably by definition the most different you can make an image ;d) is used to mark the number one priorities. Upgrades you can buy!
> I don't expect the green to distract players for long. That is, by far, the most standardized visual metaphor in the game when it comes to upgrades. I've long since learned to "filter out" green upgrades when I'm looking for what to purchase... or to _focus_ on the green upgrades when I want to review my progress toward current objectives.

And yet Plat decided to **not** use green in the singularity shop. Maybe because it's an evolution of the quark shop, which had icons so big that we needed a hide maxed as the number of upgrades grew? Either way, it's a meaningless UI change with no affect on balance, only usability. I believe it helps new players far more than it distracts seasoned players from being able to click on a memorized box. The fact that I've seen as many positive comments as negative comments about the feature makes me think it's a net gain. Considering that some of the complaints are along the lines of 'why did we get this when all we wanted was a hide max' or something that would be programmatically much harder to implement, I do give more weigh to the positive side.

> As for the changes currently in this PR, I'll admit that I didn't think about red vs green colorblindness

Neither did I, until after the fact. It was just luck that it mostly works for various visual impairments. Or subliminal.

Part of me wonders if there would be any negative reaction if this was always the way it was. I feel like it's super useful in the beginning, and by the time you don't need it as much you've already tuned it out.

Sadly, I feel like I'm the only 'discussion' you'll get around this. I tried soliciting feedback on my original PR and it was all crickets. It's great to have someone at least interested to talk to ;p My goal is/was to make it more intuitive for new players. What can we buy? Show them. Are the complaints of the seasoned veterans worth sacrificing that basic premise? I've thought about it a lot. My choices weren't willy-nilly. I'm not trying to prove you wrong, just provide my perspective so you can understand if not agree. Obviously you care enough to contribute, so that's a good thing. Collaboration is fantastic.

TL;DR: Grumpy old farts in later progression channels that don't need any guidance on which upgrades are purchasable do not get to complain about a feature that provides said guidance to new players. Stay off their lawns. (P.s. I am old. And grumpy. But not progressed!) It's late here, but you definitely deserved a response. I hope i was cogent enough; I tried to proofread, but it's very late. I hope i didn't misspell imagine again >_<

Thank you for your consideration. <3

jahwsuf:
> My bias is also to the new player, to people new to the mechanics, and the most frequent questions in discord channels. Not necessarily relevant here, but either I notice a rough edge personally, or the frequency of the same question / irk draws me to it. You plop down 61 Singularity upgrades that are not sorted by cost, are also level locked, and it's hard to tell when you can buy and when. Researches also suffer from this; but it's a more farmable resource, and then the autobuyer steps in. Wowcube upgrades also have a similar issue, but again it's another relatively quickly farmable resource. GQ are a little more precious.
> 
> You shouldn't have to be on discord to play the game, the game should be able to help guide you. In general it does, but the Sing Shop felt a little lacking. Hence the (old) PR.

I completely agree.  That is definitely a daunting wall of disorganized upgrades.

The standard set of upgrades is generally well-ordered:  left to right, top to bottom.  Same with researches, though there _is_ a little hopping.  Even then, the fact that many are hidden until certain challenges are completed explains the reasons for the discontinuities.  That 'unlock' factor for researches kept the info flood staggered and manageable - everything there was simpler at first.  It then expanded what was available as the player progressed - so it was never nearly as bad there.  (Unlock thresholds are often a positive thing in games for this very reason!)  Heck, this observation also holds for achievements - left to right, with unlock staging.

But _then_ there are the Singularity upgrades - everything is visible, all at once, that you could ever buy with golden quarks no matter what the minimum singularity level is.  Yep, help is _definitely_ needed there.  100% on board with everything quoted above.  Even though I'm not a fan of the styling, I am a fan of the "something is better than nothing" principle, so thanks for making that PR!

> And yet Plat decided to not use green in the singularity shop. Maybe because it's an evolution of the quark shop, which had icons so big that we needed a hide maxed as the number of upgrades grew?

I've noticed that the related icons lack transparency / alpha channels, unlike the icons for the older upgrades.  If those images were donated, and it's hard to add the alpha channels in after the fact... that _would_ make the green hard to implement.

Of course, the previous paragraph being the actual explanation is 90% speculation, 10% inference on my end; it may well be intentional that green wasn't used, regardless of the transparency issue. Without his explicit word on the matter, I don't think we'll know for sure.

> Sadly, I feel like I'm the only 'discussion' you'll get around this.

... I'm kind of noticing that, yeah.

jahwsuf:
Seems like concerns about the formatting have died down, and there's not much interest.  I'll go ahead and retract this.

I'll keep the branch up on my fork for a bit as a just-in-case.

PR #436, opened 2023-01-13
Endure harsh conditions

httpsnet:
We found and fixed an issue where Quark, Golden Quarks and Octeracts would become NaN due to the Twenty Ascensions Challenge.
Since NaN are set to 0 on loading, they cannot be penalized due to lack of evidence of cheating. Do not exploit.
There are also some other improvements to make it more resistant to extreme singularities or exploits. The modified code was checked many times over a day's worth of time.

Changed the way Patreon bonuses are saved. In order to protect against exploits by modification, we encoded with Base64 to increase the difficulty of modification. If there is a problem with saving, it will be corrected automatically.
Save the current version on each load. It used to work the last time you did Singularity.
calculateEffectiveSingularities() capped at 1e300. Because I catastrophically broke my system with Infinity.
Ascension speed division with Twenty Ascensions Challenge now maxes out at 1e100.
Octeracts gain is now ignored for NaN. Octeracts capped at 1e300.
Global Speed and Ascension Speed caps are now above 1e-300 and below 1e300.
Quark Shop and Singularity Upgrades and Octeracts Upgrades can no longer be purchased if the cost item is not a finite number.
Singularity Upgrades and Octeracts Upgrades can now specify the number of purchases if you enter minus 2 or less at the prompt.
Max purchase of Singularity Upgrades is now 1000000 if the highest Singularity is 250 or higher.
Fixed incorrect displayed calculations for Divine Pack and Abstract Exokinetics.


KhafraDev (changes requested):
Storing of the patreon bonus shouldn't be changed.

httpsnet:
I put it back, but the security risk is terrible

KhafraDev:
base64 encoding the value isn't much safer. Rather, the value should be stored in cache and fetched on load.

httpsnet:
I won't touch the patreon bonus this time. Prioritize resolving other issues.

httpsnet:
It's been 5 days, what happened?
This PR fixes the issue that occurred with this URL.
https://discord.com/channels/677271830838640680/1061317235727155290

NaN can buy upgrades until it hits infinity, but it can't punish exploits that use it, so if this issue gets hyped up on general channels, people's reputation for Synergism will suffer a lot.
should merge soon

src/singularity.ts
```diff
@@ -136,20 +136,31 @@ export class SingularityUpgrade extends DynamicUpgrade {
     public async buyLevel(event: MouseEvent): Promise<void> {
         let purchased = 0;
         let maxPurchasable = 1
+        player.goldenQuarks = 1e100
```
KhafraDev, line 139:
```suggestion

```

KhafraDev (changes requested):

src/singularity.ts
httpsnet: It seems that I forgot to erase the garbage in the test. Forgive me.

PR #440, opened 2023-01-15
Refactored CalcECC function to simplify calls.

Yayformee:
Changed the parameters from (type, #completions) to (index).
This can be done since both the type and the #completions strictly depend on the challenge index.


KhafraDev (approved):
Thanks for the PR! We're currently working on other things so this PR will probably be delayed a little longer.

KhafraDev (changes requested):
I'm sorry but I've changed my mind. It does simplify it, but I think it's too much if that makes sense. I like that it has to be explicitly called with a type; I'd rather keep it that way. Of course a change where moving `player.challengecompletions[index]`s could be moved to CalcECC in a way that still simplifies the code while still making the function explicit.

PR #443, opened 2023-01-16
i18n enhancement:  adds language-select subtab under Settings

jahwsuf:
The new feature:

image

A button will automatically be added for each entry in the `supported` language _map_ defined in i18n.ts.  I did convert `supported` to a 'map' / object rather than leave it as an array; this allows the properties needed for each language button to be properly specified.

From this [PR's changeset](https://github.com/jahwsuf/SynergismOfficial/blob/537bdccfebade0aaa0d5420f666062f10fb1e105/src/i18n.ts#L5-L11):

```typescript
// For 'flag': https://emojipedia.org/emoji-flag-sequence/
// Searching "flag <country>" in their search bar will help verify the code.
const supported: {[langID: string]: {name: string, flag: string}} = {
    // Define language properties and mappings here.
    en: { name: 'English', flag: '🇺🇸'}, // Or '🇺🇸 / 🇬🇧', no name?
    zh: { name: 'Chinese', flag: '🇨🇳'}
};
```

Naturally, feel free to change the information in the right-hand side as desired; I simply thought it'd help fill some of the whitespace and would also provide a good starting point for people to get involved should they care about those details.

**_Other notes_**

I made sure to style it to handle additional languages.  Via node duplication in Developer mode:

image

It's not perfect, but it should be a "good enough" starting place.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3010,7 +3010,7 @@ $STAGE$ for synergism stage"></p>
             <p>
                 Interested in providing a translation for your language?  Visit the <a href="https://www.discord.gg/ameCknq">#translations channel on our Discord!</a>
             </p>
-            <p>Visit our Crowdin project at <a href="https://crowdin.com/project/synergism-official">https://crowdin.com/project/synergism-official</a>!</p>
+            <p>Visit our POEditor project at <a href="https://poeditor.com/join/project/1up4tT24vG">https://poeditor.com/join/project/1up4tT24vG</a>!</p>
             <p>Contribute directly through Github at <a href="https://github.com/Pseudo-Corp/SynergismOfficial">https://github.com/Pseudo-Corp/SynergismOfficial</a>!</p>
         </div>
     </div>
--- a/index.html
+++ b/index.html
@@ -2866,12 +2866,13 @@
 <div id="settings" alt="settings" label="settings" style="display: none;">
     <div class="subtabSwitcher subTabWrapper">
         <button id="switchSettingSubTab1" alt="switchSettingSubTab1" label="switchSettingSubTab1" class="buttonActive" style="border: 2px solid white">Settings</button>
-        <button id="switchSettingSubTab2" alt="switchSettingSubTab2" label="switchSettingSubTab2" style="border: 2px solid gold">Credits</button>
-        <button id="switchSettingSubTab3" alt="switchSettingSubTab3" label="switchSettingSubTab3" style="border: 2px solid orange">Stats for Nerds</button>
-        <button id="switchSettingSubTab4" alt="switchSettingSubTab4" label="switchSettingSubTab4" class="prestigeunlockib" style="border: 2px solid brown">Reset History</button>
-        <button id="switchSettingSubTab5" alt="switchSettingSubTab5" label="switchSettingSubTab5" class="ascendunlockib" style="border: 2px solid darkorange">Ascend History</button>
-        <button id="switchSettingSubTab6" alt="switchSettingSubTab6" label="switchSettingSubTab6" class="singularity" style="border: 2px solid darkorange">Singularity History</button>
-        <button id="switchSettingSubTab7" alt="switchSettingSubTab7" label="switchSettingSubTab7" style="border: 2px solid white">Hotkeys</button>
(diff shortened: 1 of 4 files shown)


src/i18n.ts
```diff
@@ -1,12 +1,19 @@
 import i18next, { type Resource } from 'i18next'
 import { prod } from './Config'
+import { DOMCacheGetOrSet } from './Cache/DOM';
 
-const supported = ['en', 'zh']
+// For 'flag': https://emojipedia.org/emoji-flag-sequence/
+// Searching "flag <country>" in their search bar will help verify the code.
+const supported: {[langID: string]: {name: string, flag: string}} = {
```
KhafraDev, line 7:
```suggestion
const supported: Record<string, { name: string, flag: string }> = {
```
we want to eventually move to a stricter typescript lint

index.html
```diff
@@ -2999,6 +3000,20 @@
...
+            <div id="languageOptions" class="languageOptions">
+                <!-- Populated by i18n.ts, `buildLanguageTab` -->
+            </div>
+        </div>
+        <div class="languageContribution">
+            <p>
+                Interested in providing a translation for your language?  Visit the <a href="https://www.discord.gg/ameCknq">#translations channel on our Discord!</a>
+            </p>
+            <p>Visit our Crowdin project at <a href="https://crowdin.com/project/synergism-official">https://crowdin.com/project/synergism-official</a>!</p>
+            <p>Contribute directly through Github at <a href="https://github.com/Pseudo-Corp/SynergismOfficial">https://github.com/Pseudo-Corp/SynergismOfficial</a>!</p>
```
KhafraDev, line 3014: This text should also be translate-able. I guess having a giant method in i18n.ts that changes text in the html file on load/language change would suffice. You can add text to translations/default.json and I will upload them to crowdin.

KhafraDev (changes requested):

index.html
jahwsuf: That's totally fair.

Before I go through the trouble, is this text about what you'd hope to have for the tab?  Or would different info be preferred?

src/i18n.ts
jahwsuf: Huh... how did I never run into the `Record` type until now?  It's been around long enough, according to the docs.

jahwsuf:
It's the end of the day where I am; I'll have to pick back up with this later.  I don't mind y'all taking this to completion without me if I don't get back to it in time, but it shouldn't be more than a few days' wait at the most.

index.html
jahwsuf: What's the current planned strategy for doing this with other index.html strings, if there is one?  I probably shouldn't diverge too much for whatever that strategy may be.

Even if not everything is localized, I imagine the tab titles and subtab titles should probably be localized... and handling for that would face very similar issues.

My _first_ thought is using element IDs as i18n string keys, since they're usually unique on the page, but if you have something else in mind, feel free to suggest something more aligned to your expectations.

KhafraDev: Crowdin should be replaced with poeditor (https://poeditor.com/join/project/1up4tT24vG) now, but everything else seems fine.

KhafraDev: I could see element keys be named (or have an i18n- attribute) so we don't have to move all the text out of the html file.

```html
<div i18n="key.something.name" i18n-replace="{{name:abc}}" />
```

where the key would be
```json
{
  "key": {
    "something": {
      "name": "Your name is {{name}}"
    }
  }
}
```

although better left off to another PR

index.html
```diff
@@ -2999,6 +3000,20 @@
...
+        <div class="languagesMenu">
+            <div id="languageOptions" class="languageOptions">
+                <!-- Populated by i18n.ts, `buildLanguageTab` -->
+            </div>
+        </div>
+        <div class="languageContribution">
+            <p>
+                Interested in providing a translation for your language?  Visit the <a href="https://www.discord.gg/ameCknq">#translations channel on our Discord!</a>
+            </p>
+            <p>Visit our Crowdin project at <a href="https://crowdin.com/project/synergism-official">https://crowdin.com/project/synergism-official</a>!</p>
```
KhafraDev, line 3013:
```suggestion
            <p>Visit our POEditor project at <a href="https://poeditor.com/join/project/1up4tT24vG">https://poeditor.com/join/project/1up4tT24vG</a>!</p>
```

PR #444, opened 2023-01-17
Addition of NOT, have, have not, and (Softcapped).

Mixelz:
Cobbled together some additional generals for the translation, added the appropriate lines to the Chinese.json in English for others to translate.
Added in 2 new lines for when the player has no perks left to unlock or improve.
Finally, picked up a dropped "t" for the GQ2 "effec".

### DIFF
--- a/src/Octeracts.ts
+++ b/src/Octeracts.ts
@@ -237,7 +237,7 @@ export const octeractData: Record<keyof Player['octeractUpgrades'], IOcteractDat
         effect: (n: number) => {
             return {
                 bonus: n > 0,
-                desc: `Octo-Hepteract Primality Synergism is ${n > 0 ? '' : 'NOT'} active.`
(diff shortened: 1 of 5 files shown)


jahwsuf:
This sort of thing isn't as language-independent as you may think.

French uses two words to negate a sentence rather than one:  'n'est ... pas', if I recall right.

Khmer (from Cambodia) acts similarly - there's one word in the "normal" not position, but the other word is near-always sentence-final.

... language stuff is hard.

Mixelz:
I understand this, I said as much in the discord server but it bears repeating here that this system only works for languages that have this abilty to modify the entire sentence with a single NOT.

Ideally we'd just have two seperate sentances for if you or do not have something purchased but that'll require a lot more text for translation overall.

Mixelz:
Removing this as it's become redundant.

PR #445, opened 2023-01-19
fix: restores stats for nerds updates

jahwsuf:
Since the "Languages" subtab was added before "Credits" (and before "Stats for Nerds")... the "Stats for Nerds" subtab id got incremented. I didn't catch this when writing #443.

image

So, this is a simple fix to get the stats page auto-updating again.


jahwsuf:
Minor shout-out to @httpsnet for giving a nudge in the right direction in the Discord.

PR #469, opened 2023-04-06
feat: revamp events, make code readable

KhafraDev:
- buffs show their object key name rather than a formatted name (wasn't super high priority for me to fix)
- ~~isEvent is now a getter and probably slow, we should memoize/debounce it for perf~~

### DIFF
--- a/src/Variables.ts
+++ b/src/Variables.ts
@@ -1,5 +1,6 @@
 import Decimal from 'break_infinity.js'
 import type { GlobalVariables } from './types/Synergism'
+import { Events } from './Event'
 
 export enum Upgrade {
      coin = 'coins',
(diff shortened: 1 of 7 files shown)


src/Variables.ts
```diff
@@ -423,7 +424,12 @@ export const Globals: GlobalVariables = {
 
   historyCountMax: 20,
 
-  isEvent: false,
+  get isEvent () {
+    return Events.list.active.length > 0
+  },
+
+  set isEvent (_: boolean) {},
```
KhafraDev, line 431: funny hack because we do `Object.assign(Globals, { ...blankGlobals })` which can't be done on a readonly property

marland34562 (approved):
its gonna be cool!

PR #505, opened 2023-05-26
The B I G Perk UI Overhaul

Mixelz:
After Thousands of Years, the hour is at hand! Rejoice and celebrate for I have FINALLY figured out how the hell to implement the majority of my crazy changes for a full overhaul of the Perk Ui Menu.

**Changes:**

- Complete Overhaul on Perk UI; all perks now display their description, level and the Sing they were given/upgraded on in a dedicated description box!
- All Perks have an icon as well, art done by myself! (Intended to be used with the Legacy Icon Pack)
- Merged the Penalties tab into the Perk Tab due to the non-utlization of the Penalties tab space. (It was super empty.)
- Backend changes to accomodate the above, mostly located in singularity.ts (Perks now have an internal ID, using their .json name for now.)

Extensive care was taken to ensure the overhaul works for all currently supported languages and should be future proof for any additional ones that may be implemented down the road.

I'd also like to give thanks to awWhy and sasha for their massive help in walking me through how to get description hover capabilities setup. Thank you both so much for the help!

**Preview:**
image


Mixelz:
There should only be one additional PR I'll need to make before I'm done with the perk Project, which would be remaking & reformatting all current perk descriptions to better utilize the new perk description box.

That itself is gonna require a lot of time and a lot of work, so I am saving that for another day. 
For now everything is fine to be pushed as implemented. :)

KhafraDev (changes requested):
The subtab changes don't seem correct to me. Could you explain them?

https://github.com/Pseudo-Corp/SynergismOfficial/blob/f4baffa7a2c9662c85b7b64296e3ae4a44f75655/index.html#L3912

Mixelz:
> The subtab changes don't seem correct to me. Could you explain them?
> 
> https://github.com/Pseudo-Corp/SynergismOfficial/blob/f4baffa7a2c9662c85b7b64296e3ae4a44f75655/index.html#L3912

The subtab changes were done due to consolodating the Perks & Penalties tabs together. 
Everything was indexed from them being two seperate additions to the list, so had to increment all tabs past them down by one where applicable, also accounting for the small value changes scattered throughout toggle.ts, UpdateHTML.ts and UpdateVisuals.ts.

src/singularity.ts
```diff
@@ -1622,34 +1657,33 @@ export const singularityPerks: SingularityPerk[] = [
...
-
-  const str = `${getAvailablePerksDescription(singularityCount)}`
 
   DOMCacheGetOrSet('singularityPerksHeader').innerHTML = strH
-  DOMCacheGetOrSet('singularityPerksDesc').innerHTML = strD
-  DOMCacheGetOrSet('singularityPerksGrid').innerHTML = str
+  DOMCacheGetOrSet('singularityPerksText').innerHTML = `${i18next.t('general.level')} # - (Singularity #)`
+  DOMCacheGetOrSet('singularityPerksDesc').innerHTML = i18next.t('singularity.perks.description')
+  DOMCacheGetOrSet('singularityPerksGrid').innerHTML = `${getAvailablePerksDescription(singularityCount)}`
+  addPerkEventHandlers()
```
KhafraDev, line 1686: if `updateSingularityPerks` is called more than once this will create a memory leak.

src/singularity.ts
```diff
@@ -1734,11 +1769,27 @@ const getAvailablePerksDescription = (singularityCount: number): string => {
...
+  return `<span${isNew?' class="newPerk"':' class="oldPerk"'} id="${perkData.ID}"> 
+  <img src="Pictures/Default/perk${perkData.ID}.png" loading="lazy">
+  ${perkData.name}
+  </span>`
 }
 
+// Attach EventListener to perks to allow for on hover capabilities.
+export function addPerkEventHandlers() {
+  for (const perk of singularityPerks) {
+    document.getElementById(perk.ID)!.addEventListener('mouseover', () => {
```
KhafraDev, line 1782: is there a reason to use document.getElementById?

src/singularity.ts
```diff
@@ -1734,11 +1769,27 @@ const getAvailablePerksDescription = (singularityCount: number): string => {
...
+      const currentLevel = getLastUpgradeInfo(perk, player.highestSingularityCount).level
+      const lastUpgraded = getLastUpgradeInfo(perk, player.highestSingularityCount).singularity
+      const levelInfo = `${i18next.t('general.level')} ${currentLevel} - (Singularity ${lastUpgraded})`
+      const perkImage = DOMCacheGetOrSet('singularityPerksIcon') as HTMLImageElement
+      perkImage.src = `Pictures/Default/perk${perk.ID}.png`
+      DOMCacheGetOrSet('singularityPerksText').textContent = `${levelInfo}`
```
KhafraDev, line 1788:
```suggestion
      DOMCacheGetOrSet('singularityPerksText').textContent = levelInfo
```

levelInfo is already a string

src/singularity.ts
```diff
@@ -1734,11 +1769,27 @@ const getAvailablePerksDescription = (singularityCount: number): string => {
...
+      DOMCacheGetOrSet('singularityPerksDesc').textContent = `${perk.description(player.highestSingularityCount, perk.levels)}`
```
KhafraDev, line 1789:
```suggestion
      DOMCacheGetOrSet('singularityPerksDesc').textContent = perk.description(player.highestSingularityCount, perk.levels)
```

same here

translations/de.json
```diff
@@ -12,7 +12,7 @@
     "ambrosia": {
         "amount": "You have <<var(--lightgoldenrodyellow-text-color)|{{ambrosia}}>> Ambrosia!",
         "blueberryGeneration": "Each second, gain <<var(--amber-text-color)|{{chance}}>> seconds of Blueberry Time.",
-        "perGen": "Currently gaining <<var(--amber-text-color)|{{guaranteed}}>> with a <<var(--amber-text-color)|{{extraChance}}%>> chance of another. [<<var(--lime-text-color)|Luck: &#9752 {{ambrosiaLuck}}>>]",
```
KhafraDev, line 15: I'm not sure about these changes. I assume crowdin will revert them.

KhafraDev (changes requested):

translations/de.json
Mixelz: The clover change was made due to the way Perk Descriptions are handled, it does not properly read out the Decimal code and instead displays it raw. 

For perks it will be a required change in order for the text to display properly, I made the change globally to help conform it across the board and due to it being easier to read than the raw decimal code itself. I do not believe there should be any negative effects from this.

src/singularity.ts
Mixelz: This is required due to caching, with a standard DOMCacheGetOrSet implementation the description box breaks whenever the tab has to reload due to switching to another primary tab. 
The discussion in the progamming channel on the discord server goes into a bit more detail about it. https://discord.com/channels/677271830838640680/705896428287033375/1111390816208953437

src/singularity.ts
Mixelz: Part of the issue is that the perk menu seems to recreate itself whenever the Singularity tab is visited, where abouts would be a better place to put this function given this?

awWhy: As long as this function is called only after any sets to 'singularityPerksGrid' innerHTML, shoudn't be any memory leaks (that if we assume everyone uses modern browsers that garbage collects no longer connected events), that's also another reason for not using DOMCacheGetOrSet, we want to address new element with same ID, while DOM only allows to address old one (also would cause memory leak since events would pile up on top of each other)

translations/de.json
KhafraDev: It's unlikely that Platonic added these in himself, which means crowdin likely did so.

Mixelz: I'll throw them in for now and if anything besides the perk text gets changed then it'll be no big deal.

KhafraDev (commented):
My biggest issue is not the code itself, but how it's designed. Recreating the elements on every refresh & relying on garbage collection to prevent a likely candidate for a memory leak just doesn't sit right with me. Garbage collection in js is nondeterministic and can be slow. I'd recommend reading the node docs about it [^1]. If you were to trace the garbage collection of Synergism, you'd see few mark-sweep gcs and a bunch of scavenge (which is good). With this change you'd undoubtedly see 1) more allocations and 2) more mark-sweep garbage collections. While it's unlikely to cause issues, simply due to Synergism being efficient *enough*, I'd rather not add in a mechanic that, simply put, is capable of causing memory leaks.

Rather, the top level elements should only be created once & event listeners only appended once, and then the element's children should be modified instead of recreated. Then you can rely on the existing caching and we don't have to worry about possible memory leaks. Win-win. This is what EventListeners.ts does: event handlers are added once on load, and then never done again.

[^1]: https://nodejs.org/en/docs/guides/diagnostics/memory/using-gc-traces

Mixelz:
No worries, read through the comment and while I understand it does leave me in a bit of a pickle...

Mostly due to how the perk menu is currently implemented. From my understanding it creates itself whenever the tab is loaded or reloaded in, I am unsure of the reason why but if I had to guess it might be due to the fact that it has the capability to change its total contents based on player progression, though it could be said that the quark shop also handles itself in a similar way so I truthfully have no idea beyond it is how it is.

As it currently stands I'd need to more thoroughly dismantle and overhaul the way perks work on the backend to accomodate the review which will certainly put a dent in my ability to actually carry it out. I suppose I am going to try my best in that regard and see what I can actually do, here's hoping I can pull something out of it.

awWhy:
To create static list that updates its own order without any innerHTML, is really hard to explain
Way easier to find a way to remove event listeners before innerHTML assignment (as any innerHTML are already going to garbage collector no matter what)

awWhy:
Maybe instead of changing how it works, better to make perks update only once in Singularity reset (or save file load)
As creating static HTML would require a lot of maintance (that Plat will not remember to do)
Removing event listeners manually requires non anynomous functions to be used (which requires storaging them or adding new dependicy)

Mixelz:
Submitted the rewrite of the Event Listeners, hopefully it is up standards, if not please let awWhy and I know what will need to be changed.

src/EventListeners.ts
```diff
@@ -46,7 +46,7 @@ import { changeTab, changeSubTab } from './Tabs'
     Platonic and/or Khafra have the right to close PRs that do not conform to this style guide
 
     If you are editing this script, please update the below time:
-    Last Edited: June 10, 2021 3:04 AM UTC-8
+    Last Edited: Sunday, 28-May-23 22:36:58 UTC
```
KhafraDev, line 49: for future reference I don't know what Platonic was going for here but you don't need to update comments like these

KhafraDev (approved):

Pseudonian:
Thank you so much for the PR! Looks great, fantastic job.

PR #511, opened 2023-06-09
Blueberry Tree!

Pseudonian:


src/BlueberryUpgrades.ts
```diff
@@ -0,0 +1,444 @@
...
+
+      preReqText = preReqText.slice(0, -1)
+    }
+
+    return `<span style="color: gold">${this.name}</span>
+                ${preReqText}
+                <span style="color: lightblue">${this.description}</span>
+                <span style="color: ${color}"> ${i18next.t('general.level')} ${format(this.level, 0, true)}${maxLevel}${freeLevelInfo}</span>
+                <span style="color: gold">${this.rewardDesc}</span>
+                ${i18next.t('octeract.toString.costNextLevel')}: <span style="color:orange">${format(costNextLevel, 2, true, true, true)}</span> Ambrosia ${affordableInfo}
```
KhafraDev, line 166: "Ambrosia" should use i18n, if it's not already translated

src/BlueberryUpgrades.ts
```diff
@@ -0,0 +1,444 @@
...
+  ambrosiaQuarkCube1: {
+    maxLevel: 25,
+    costPerLevel: 250,
+    blueberryCost: 1,
+    costFormula: (level: number, baseCost: number): number => {
+      return baseCost * (Math.pow(level + 1, 3) - Math.pow(level, 3))
+    },
+    rewards: (n: number) => {
+      const baseVal = 0.0005 * n
+      function val(base: number) {
```
KhafraDev, line 300: I don't see a case where val is needed (same goes for all below upgrades too), it's being called twice with the same value every time. It could be simplified to

```js
const val = 1 + baseVal * Math.floor(Math.pow(Math.log10(Number(player.worlds)+1) +1, 2))
```

(etc.) for the ones below

src/BlueberryUpgrades.ts
```diff
@@ -0,0 +1,444 @@
...
+      'ambrosiaLuck1': 30,
+      'ambrosiaQuarks1': 20
+    },
+    cacheUpdates: [() => player.caches.ambrosiaLuck.updateVal('BlueberryQuarkLuck1')]
+  }
+
+}
+
+export const resetBlueberryTree = () => {
+  for (const upgrade in player.blueberryUpgrades) {
```
KhafraDev, line 439:
```suggestion
  for (const upgrade of Object.keys(player.blueberryUpgrades)) {
```

src/Calculate.ts
```diff
@@ -2178,41 +2199,47 @@ export const calculateAmbrosiaLuckShopUpgrade = () => {
...
 
-  if (player.lifetimeAmbrosia >= 1e7) {
-    val *= 2
-  }
+export const calculateSingularityMilestoneBlueberries = () => {
+  let val = 0
+  if (player.highestSingularityCount >= 66) val += 1
+  if (player.highestSingularityCount >= 132) val += 1
+  if (player.highestSingularityCount >= 198) val += 1
+  if (player.highestSingularityCount >= 254) val += 1
```
KhafraDev, line 2221:
```suggestion
  if (player.highestSingularityCount >= 254) val = 4
  else if (player.highestSingularityCount >= 198) val = 3
  else if (player.highestSingularityCount >= 132) val = 2
  else if (player.highestSingularityCount >= 66) val = 1

```

?

src/CheckVariables.ts
```diff
@@ -750,6 +766,38 @@ export const checkVariablesOnLoad = (data: PlayerSave) => {
     }
   }
 
+  if (data.blueberryUpgrades != null) {
+    for (const item in blankSave.blueberryUpgrades) {
```
KhafraDev, line 770:
```suggestion
    for (const item of Object.keys(blankSave.blueberryUpgrades)) {
```

KhafraDev (changes requested):
some pretty minor nits

PR #519, opened 2023-06-23
code 'add': Allow saving of attempts and prefill using all attempts

hex7cd:
This PR adds 3 separate things:

1. The functions Prompt and PromptCB (from UpdateHTML.ts) can now use an optinal third parameter of type string to prefill the input field.
2.  code 'add' uses this to prefill the field with all available attempts, which I believe is what most people use it for anyway.
3. Players can enter a negative amount of attempts into the code 'add' prompt to save that many attempts (same as with opening cubes).

Images:
code add prefill
code add negative entry
code add negative entry 2

### DIFF
--- a/src/ImportExport.ts
+++ b/src/ImportExport.ts
@@ -491,7 +491,7 @@ export const promocodes = async (input: string | null, amount?: number) => {
     if (amount) {
       attemptsUsed = amount.toString()
     } else {
-      attemptsUsed = await Prompt(i18next.t('importexport.useXAdds', { x: availableUses }))
+      attemptsUsed = await Prompt(i18next.t('importexport.useXAdds', { x: availableUses }), availableUses.toString())
     }
 
     if (attemptsUsed === null) {
--- a/src/ImportExport.ts
+++ b/src/ImportExport.ts
@@ -501,14 +501,15 @@ export const promocodes = async (input: string | null, amount?: number) => {
     if (
       Number.isNaN(toUse) ||
             !Number.isInteger(toUse) ||
-            toUse <= 0
+            toUse === 0 ||
+            (toUse < 0 && -toUse >= availableUses)
     ) {
       return Alert(i18next.t('general.validation.invalidNumber'))
     }
 
     const addEffects = addCodeBonuses()
 
-    const realAttemptsUsed = Math.min(availableUses, toUse)
+    const realAttemptsUsed = toUse > 0 ? Math.min(availableUses, toUse) : availableUses + toUse
     const actualQuarks = Math.floor(addEffects.quarks * realAttemptsUsed)
     const [first, second] = window.crypto.getRandomValues(new Uint8Array(2))
 
--- a/src/UpdateHTML.ts
+++ b/src/UpdateHTML.ts
@@ -1074,7 +1074,7 @@ const AlertCB = (text: string, cb: (value: undefined) => void) => {
     popup.addEventListener('keyup', kbListener)
 }
 
-export const PromptCB = (text: string, cb: (value: string | null) => void, prefill?: string) => {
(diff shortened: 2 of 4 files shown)


KhafraDev (changes requested):

src/UpdateHTML.ts
```diff
@@ -1074,7 +1074,7 @@ const AlertCB = (text: string, cb: (value: undefined) => void) => {
     popup.addEventListener('keyup', kbListener)
 }
 
-export const PromptCB = (text: string, cb: (value: string | null) => void) => {
+export const PromptCB = (text: string, cb: (value: string | null) => void, prefill?: string) => {
```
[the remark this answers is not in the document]
hex7cd: I have implemented your suggestion.

src/UpdateHTML.ts
```diff
@@ -1089,6 +1089,9 @@ export const PromptCB = (text: string, cb: (value: string | null) => void) => {
   confWrap.style.display = 'block'
   overlay.style.display = 'block'
     popup.querySelector('label')!.textContent = text
+    if (typeof(prefill) !== 'undefined') {
```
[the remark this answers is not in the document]
hex7cd: You are right, empty string should not serve any purpose, so I implemented your second suggestion.

translations/en.json
```diff
@@ -3216,7 +3216,7 @@
     "promocodePrompt": "Got a code? Great! Enter it in (CaSe SeNsItIvE). \n [Note to viewer: this is for events and certain always-active codes. \n May I suggest you type in \"synergism2021\" or \"Khafra\" perchance?]",
     "comeBackSoon": "Alright, come back soon!",
     "noAddCodes": "You do not have an 'Add' code attempt! You will gain 1 in {{x}} seconds.",
-    "useXAdds": "You can use up to {{x}} attempts at once. How many would you like to use?",
+    "useXAdds": "You can use up to {{x}} attempts at once. How many would you like to use? You can start the input with \"-\" (put in a negative value) to put in a value, that will be saved, which means you use all attempts except the specified number.",
```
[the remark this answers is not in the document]
hex7cd: Should I make a PR for the same change for cubes?

src/ImportExport.ts
```diff
@@ -501,14 +501,15 @@ export const promocodes = async (input: string | null, amount?: number) => {
     if (
       Number.isNaN(toUse) ||
             !Number.isInteger(toUse) ||
-            toUse <= 0
+            toUse === 0 ||
+            (toUse < 0 && -toUse >= availableUses)
```
KhafraDev, line 505: if someone with 10 add uses puts in -10 it won't work, I don't understand the restriction. Not a major issue but I'd like to know the reasoning

hex7cd: Using zero attempts was forbidden before this PR (toUse <= 0) and I didn't see a reason to change that. I wrote the validity check with that in mind.

hex7cd: Wouldn't this simplification also be possible? Should check for both zero and negative values at once.
```suggestion
Number.isNaN(toUse) ||
    !Number.isInteger(toUse) ||
    -toUse >= availableUses
```

KhafraDev: I'm referring to someone entering -10 to use all of their hypothetical 10 add uses, which wouldn't work because -(-10) <= 10

hex7cd: If I enter -10, that means I want to use all but 10 uses. If I only have 10 uses and enter -10, that means I want to save 10 out of 10 uses and use effectively none of them. If I want to use all of my 10 add uses, I'd put in 10 without a minus sign (which is also what the field comes pre-filled with then).

KhafraDev: that makes sense, wording confused me

PR #524, opened 2023-06-29
feat: adds loadout highlighting and description on import

jahwsuf:
I thought I'd take a bit of time to add some quality-of-life for the new loadout feature.

1. When importing a loadout from a file, the game will display the loadout's settings afterward:

image

2. When hovering over a blueberry loadout button, any upgrades purchased within that loadout will be highlighted:

image

image

The highlighting effect will disappear on 'mouseout'.

3. From the Discord:

    image

    I was looking at the function anyway for this, so I threw in the fix.

### DIFF
--- a/Synergism.css
+++ b/Synergism.css
@@ -3714,3 +3714,14 @@ img#singularityPerksIcon {
     max-width: 500%;
     animation: rotation 2s infinite linear;
 }
+
+.hoveredBlueberryLoadout1 div.bbPurchasedLoadout1,
+.hoveredBlueberryLoadout2 div.bbPurchasedLoadout2,
+.hoveredBlueberryLoadout3 div.bbPurchasedLoadout3,
+.hoveredBlueberryLoadout4 div.bbPurchasedLoadout4,
+.hoveredBlueberryLoadout5 div.bbPurchasedLoadout5,
+.hoveredBlueberryLoadout6 div.bbPurchasedLoadout6,
+.hoveredBlueberryLoadout7 div.bbPurchasedLoadout7,
+.hoveredBlueberryLoadout8 div.bbPurchasedLoadout8 {
+    background-color: green;
+}
--- a/src/BlueberryUpgrades.ts
+++ b/src/BlueberryUpgrades.ts
@@ -633,6 +633,8 @@ export const importBlueberryTree = async (input: string | null) => {
     try {
       const modules = JSON.parse(input) as BlueberryOpt
       await createBlueberryTree(modules)
+      createLoadoutDescription(0, modules)
+
     } catch (err) {
       return Alert(i18next.t('ambrosia.importTree.error'))
     }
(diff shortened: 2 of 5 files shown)


src/BlueberryUpgrades.ts
```diff
@@ -648,6 +650,26 @@ export const loadoutHandler = async (n: number, modules: BlueberryOpt) => {
   }
 }
 
+export const updateLoadoutHoverClasses = () => {
+  const upgradeNames = Object.keys(blueberryUpgradeData) as blueberryUpgradeNames[]
+
+  for (const loadoutKey in player.blueberryLoadouts) {
```
KhafraDev, line 656:
```suggestion
  for (const loadoutKey of Object.keys(player.blueberryLoadouts)) {
```

src/BlueberryUpgrades.ts
```diff
@@ -658,12 +680,16 @@ export const saveBlueberryTree = async (input: number, previous: BlueberryOpt) =
...
   createLoadoutDescription(input, player.blueberryLoadouts[input])
+
+  updateLoadoutHoverClasses()
 }
 
 export const createLoadoutDescription = (input: number, modules: BlueberryOpt) => {
 
   let str = ''
   for (const [key, val] of Object.entries(modules)) {
+    if (!val) continue
```
KhafraDev, line 691: when is this possible? Leaving a comment would be nice

src/BlueberryUpgrades.ts
```diff
@@ -672,6 +698,11 @@ export const createLoadoutDescription = (input: number, modules: BlueberryOpt) =
   if (Object.keys(modules).length === 0) {
     str = i18next.t('ambrosia.loadouts.none')
   }
-  DOMCacheGetOrSet('singularityAmbrosiaMultiline').innerHTML = ` ${i18next.t('ambrosia.loadouts.loadout')} ${input}
+
+  let loadoutTitle = `${i18next.t('ambrosia.loadouts.loadout')} ${input}`
+  if (input == 0) {
```
KhafraDev, line 703:
```suggestion
  if (input === 0) {
```

src/BlueberryUpgrades.ts
```diff
@@ -672,6 +698,11 @@ export const createLoadoutDescription = (input: number, modules: BlueberryOpt) =
...
+    loadoutTitle = `${i18next.t('ambrosia.loadouts.imported')}`
```
KhafraDev, line 704:
```suggestion
    loadoutTitle = i18next.t('ambrosia.loadouts.imported')
```

KhafraDev (changes requested):

src/BlueberryUpgrades.ts
jahwsuf: Comment added.    tl,dr:  Point 3 in the description.

Synergism.css
```diff
@@ -3714,3 +3714,14 @@ img#singularityPerksIcon {
     max-width: 500%;
     animation: rotation 2s infinite linear;
 }
+
+.hoveredBlueberryLoadout1 div.bbPurchasedLoadout1,
```
KhafraDev, line 3718: it's probably possible to simplify this entire block of css to
```css
div[class^="hoveredBlueberryLoadout"] div[class^="bbPurchasedLoadout"] {
```

jahwsuf: It'll match the indices if we do that?  If it matches `hoveredBlueberryLoadout1` with `bbPurchasedLoadout2`, the highlighting effect between loadouts will get mixed.

jahwsuf: Further experiment:  it's actually not matching _at all_ for the highlighting, even if I drop one of the two.  Maybe it's because neither is first in the class list?

PR #531, opened 2023-07-03
Fixed Metaphysics Talisman effect not displaying.

Yayformee:


hex7cd (approved):
With this PR, Metaphysics effect shows correctly as +0.xx instead of +0
image

PR #545, opened 2023-07-10
Art stuff

LucidCrux:
Updated the art. Think everything should work fine. Changed some file names, but you might want to change them back, I don't know. The perks seem to work fine as is, but couldn't find were to change capitalization stuff.

Added new art for ambrosia luck, and blueberry time, but didn't mess with actual code since it is much more complicated now and I don't want to put the effort to learn it enough to change it, right now. But here is how I imagined it if anyone wants to implement it.
example ambrosia


KhafraDev (approved):
Thank you for your contributions again, amazing art as always! I will merge tomorrow if @Pseudonian doesn't respond.

Pseudonian:
Absolutely beautiful, once again. Thank you very much for the images!

PR #558, opened 2023-07-30
Hidden Perk Info + Mini Rework

Mixelz:
It's that time of the year again; Mixelz has another stupidly large PR for the Perks system! 🎉 

This PR was manually double checked in all languages to ensure that nothing should break, just ensure Crowdin doesn't try to overwrite anything that was CHANGED and we should be good.
This also is a bit of a experiment for Crowdin; a lot of the backend got switched around for the .json files, I'm wondering if it'll fight these changes or assume its entirely new additions and thus leave it alone. We'll find out!
### Additions
- Two new perks: Welcome to Singualrity and Automatic Sweep-o-matic Mk 2.
  - This is meant to encapsulate some additional "hidden" perks and consolidate more of their information together.
  - Welcome to Singularity is an "intro" perk that details hepteract autocraft & plat upgrade buymax. This is subject to change if we can figure out a better home for the plat upgrade buymax effect.
  - Sweep-o-matic details all challenge sweep related boosts, specifically starting with c10 & ascension challenge sweep with IC2.
- Golden Coins now displays the exact value it boosts coins by.
- TL Text for the Term "Singularity".
  - Currently used in the penalty header & perk info bar.
### Changes
- Changed the Legacy Perk Border.
  - Credits to Calmcat for the design!
- Reworked Eternal Ascensions & Ant Gods Cornucopia.
  - Eternal Ascensions now notes the Sing count mult to Asc count and moved the sweeper level into the sweeper perk
  - Ant God now mentions the 'uncorruptable' x4.44 mult, this is SEPERATE from its prior true uncorruptable bonus.
- Re-ordered specific perks on the backend.
  - This includes in the .ts and .json files, for consistency. This had to be done to get certain perks to appear at specific sections of the list.
### Removals
- Removed the description text for Singularity Achievements.
  - This is already all covered in the perks and is now redundant information.
- Removed Hepteract Autocraft perk.
  - Bankrolled it's effect into Welcome to Singularity, kept its related assets incase Plat wants it back for any reason.
### Fixes
- Automation Upgrade Level 5 is now actually awarded at S100 instead of S101.
- Fixed German Language not properly displaying Penalty numbers.
  -  Due to an extra space between each >>, go figure.

It is now 4:30 AM, I must _**S L E E P.**_ Will check on this in the morning and fix any potential issues accordingly.


src/singularity.ts
```diff
@@ -28,7 +28,7 @@ export const updateSingularityPenalties = (): void => {
 }
 
 function getSingularityOridnalText(singularityCount: number): string {
-  return `${i18next.t('general.youAreInThe')} <span style="color: gold">${toOrdinal(singularityCount)} Singularity</span>`
+  return `${i18next.t('general.youAreInThe')} <span style="color: gold">${toOrdinal(singularityCount)} ${i18next.t('general.singularity')}</span>`
```
KhafraDev, line 31: this won't be correct in all languages. We'd have to move this entire line to a singular i18n line.

src/EventListeners.ts
```diff
@@ -654,7 +654,7 @@ TODO: Fix this entire tab it's utter shit
     DOMCacheGetOrSet('singularityPerksGrid').append(perkHTML)
     DOMCacheGetOrSet(perk.ID).addEventListener('mouseover', () => {
       const perkInfo = getLastUpgradeInfo(perk, player.highestSingularityCount)
-      const levelInfo = `${i18next.t('general.level')} ${perkInfo.level} - (Singularity ${perkInfo.singularity})`
+      const levelInfo = `${i18next.t('general.level')} ${perkInfo.level} - (${i18next.t('general.singularity')} ${perkInfo.singularity})`
```
KhafraDev, line 657: this needs to be a singular translation too

KhafraDev (changes requested):
looks good so far, just these 2 comments

src/singularity.ts
Mixelz: Pushed a commit to fix this!

src/EventListeners.ts
Mixelz: Ditto to above comment

PR #560, opened 2023-07-30
various bug fixes

HypertextEye:
fixed bug: Spent Octeracts resets to 0 on singularity

fixed bug: constant upgrades 9 and 10 do not highlight when the first level is buyable

fixed bug: constant upgrades 9 and 10 are buyable after the first level, leading players to believe that buying more than one level has some additional effect

new behavior: if constant upgrade 9 or 10 are level 1+, they will not highlight, and will show a cost of 0 constant. current level is maintained, but level will be hardcapped to 1 in future singularities


KhafraDev:
thank you

PR #567, opened 2023-10-03
fix late game offerings

HypertextEye:
Offerings bugfix caused problems for late game players in early ascensions. this will resolve those problems without reverting the fix.
An alternative would be to revert the fix, please let me know if that would be preferred.


HypertextEye:
converted to draft, i'm working on a more satisfying solution to my mind

HypertextEye:
looks like I jumped the gun a bit, as this would not fix it for everyone. converted to draft again

HypertextEye:
Setting a lower bound on the time multiplier doesn't solve the issue for people who don't have absurd offering multipliers from the singularity shop (like I have). Instead I am submitting my final version (for real this time), setting a lower bound on the base offerings. I won't touch it again unless you tell me to revert everything.

HypertextEye:
Some discussion and further testing led to the conclusion that the best thing to do for now is just revert the bugfix. So that's what this PR has become. please merge, thanks

PR #568, opened 2023-11-16
bug fixes

httpsnet:
We have minimized the amount of code that needs to be modified in order to not interfere with development
Fixed 2 exploitable bugs

### minor fixes

src/Calculate.ts
Fixed Ascension Speed to be 0 if it becomes a non-numeric value
I know how to make this a NaN and it broke the game balance so I won't reveal how to do it

src/Calculate.ts
Each cube gain per ascension capped at 1e300

src/CheckVariables.ts
Fixed an issue where Singularity Upgrades could sometimes be refunded

src/UpdateHTML.ts
Fixed each cube in Ascension Stats displaying NaN when Ascension Time is 0

src/singularity.ts
Fixed an issue where the Singularity penalty could be less than 0
Currently, the only conditions for this to be less than 0 are if you have not achieved No Singularity Upgrades and purchased the Singularity Penalty Debuff

### rune fixes

src/Calculate.ts
Fixed lookupTableRuneExp referencing non-existent array

src/Runes.ts
Fixed an issue where Rune levels would not increase in some cases

### auto sacrifice fixes

src/Reset.ts
Run Rune Sacrifice and Ant Sacrifice after 0.1 seconds when ascension resets

src/Runes.ts
Fixed rune descriptions automatically referencing the last rune
​


KhafraDev (approved):

KhafraDev:
@httpsnet if you rebase I will merge.

PR #591, opened 2024-04-21
Wait for login and event processing on loading

httpsnet:
This PR will now wait for login and event handling when loading
Fetching does not occur on singularities and imports
cacheReinitialize() and dailyResetCheck() are executed after login and events

Current specifications run offline bonus without waiting for login and event when loading
As a result, ambrosia, quark, etc. were calculated without events and logins.
This is very annoying because you have to be online all the time to get the full resources.

I think it would get noisy if we discussed this issue on discord, so I'll leave it to the dev discretion
This is especially bad news for people who have used Patreon, so it should be resolved

I'm not that confident in writing the correct code, so I'll wait for approval

### DIFF
--- a/src/Event.ts
+++ b/src/Event.ts
@@ -73,6 +73,8 @@ let nowEvent: EventData | null = null
 
 export const getEvent = () => nowEvent
 
+let cacheEvents: EventData[] | null = null
+
 export const eventCheck = async () => {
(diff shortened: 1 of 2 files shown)


httpsnet:
This PR is partially inherited by #604. It is also old and will be closed

PR #594, opened 2024-05-09
Exalt UI Overhaul

Mixelz:
Oh boy another Mixelz UI Overhaul on a Singularity mechanic! **What are the odds!!!**
Changes were implemented to be as unobtrusive as possible, only 2 additional variables for EXALT challenges are required for this.
Ideally we'd also showcase current completion # under the icons like normal challenges & current effect % as well but I am not a good enough programming wizard to implement this. If anyone else wants to take a shot please do so.

## Primary Changes
- EXALT is now a Subtab of Challenges, subtabs are hidden until Exalts are unlocked normally (Sing 25).
  - Above change includes removing EXALT subtab from Singularity Tab and shifting everything required over by one (I hopefully learned my lesson from the Perk/Penalties merge).
- EXALT Rewards are now visually split between **Scaling Rewards** and **Unique Rewards**.
  - Above change includes having an bulletpoint list for all rewards in an exalt (Much easier to read than one run on sentence).
- EXALT Icons now react to the cursor (Cursor Change + Icon Highlight. WOW!!!!1!)
- EXALT Info reorganized for easier reading.

image

PR #596, opened 2024-05-31
Subtab Improvements

httpsnet:
Fixed an issue where subtab reset was not reset properly.
For example, if an unlocked subtab was displayed and a subtab was locked by import, the locked subtab could be displayed.

Subtab display state is now updated when the main tab is changed.

The argument of tabSwitcher() has been changed from a number to subTabID.
This function is used to update the visual of the subtab.

The ID name of the element of the singularity subtab has been changed to prevent the recurrence of bug #594.
Singularity has a lot of WIP, so this is to reduce the risk of index movement due to future tab movements or changes.

The tab switching function has been changed from an element style to a class.
The tab container is left as it is to avoid issues.

Fixed a case where elements in the Rune subtab were not updated.

Fixed some minor bugs in subtabs.

The tab switching function that was hooked to tabSwitcher() has been reorganized and moved to updateSubTab().

The challenge display is now updated when the C15 score is updated.
Since there were cases where the C15 score continued to be updated, a small amount of score amount required for score update was added.


httpsnet:
**Fixes**

Tab reset in resetGame() has been removed. It is executed by importSynergism()

changeTabColor(), which was only used on tab clicks, has been moved to changeTab()

Opera has been updated to fix the bug, so the workaround code has been removed

console.error(), which was used for debugging, has been removed

TabRow.constructor has been removed
this.style.cssText is already used in CSS
id declaration has been added to instant

httpsnet:
**Fixes**

Tab reset in resetGame() has been removed. It is executed by importSynergism()

changeTabColor(), which was only used on tab clicks, has been moved to changeTab()

Opera has been updated to fix the bug, so the workaround code has been removed

console.error(), which was used for debugging, has been removed

TabRow.constructor has been removed
this.style.cssText is already used in CSS
id declaration has been added to instant


This completes my work. If there is anything I can do, I will respond immediately.
If possible, please merge before the zod update

httpsnet:
This PR conflicted with other new tab improvements to the point that it was difficult to fix, so I'm closing it

PR #598, opened 2024-06-03
do not duplicate tabs in Opera

httpsnet:
We're not bad at all


httpsnet:
This is closed because it has been determined that Opera needs to be fixed

PR #611, opened 2024-07-15
Fix Exalt 5 Bonus Reward

kimihiro64:
fix global currentSingChallenge value after reloads

### DIFF
--- a/src/Synergism.ts
+++ b/src/Synergism.ts
@@ -1619,7 +1619,7 @@ const loadSynergy = async () => {
     // fix current sing challenge blank
     if (player.insideSingularityChallenge) {
       let challenges = Object.keys(player.singularityChallenges);
-      for (var i = 0; i < challenges.length(); i++) {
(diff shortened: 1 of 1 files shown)


kimihiro64:
The exalt bonus is properly applied when enabling an exalt challenge as G.currentSingChallenge is properly set, but if the game reloads, all the G values revert to the defaults, which changes it back to undefined, incorrectly calculating the bonus as 0. This repopulates the value correctly, but could just as easily not use this value and rely on the player values directly. This fix seems like less change overall in short term.

src/Synergism.ts
```diff
@@ -1616,6 +1616,17 @@ const loadSynergy = async () => {
       }
     }
 
+    // fix current sing challenge blank
```
KhafraDev, line 1619:
```suggestion
    // TODO(@KhafraDev): remove G.currentSingChallenge
    // fix current sing challenge blank
```

PR #630, opened 2025-01-12
Update singularity.ts

ZanderHill:
PATCH: Lines 2648-2649, now includes singularity debuffer effect.

### DIFF
--- a/src/singularity.ts
+++ b/src/singularity.ts
@@ -2645,8 +2645,8 @@ export const calculateSingularityDebuff = (
       ? 1 + Math.sqrt(effectiveSingularities) / 5
       : 1 + Math.pow(effectiveSingularities, 0.75) / 10000
   } else if (debuff === 'Cubes') {
-    const extraMult = player.singularityCount > 100
-      ? Math.pow(1.02, player.singularityCount - 100)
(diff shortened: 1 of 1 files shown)


src/singularity.ts
```diff
@@ -2645,8 +2645,8 @@ export const calculateSingularityDebuff = (
       ? 1 + Math.sqrt(effectiveSingularities) / 5
       : 1 + Math.pow(effectiveSingularities, 0.75) / 10000
   } else if (debuff === 'Cubes') {
-    const extraMult = player.singularityCount > 100
-      ? Math.pow(1.02, player.singularityCount - 100)
+    const extraMult = player.constitutiveSingularityCount > 100
+      ? Math.pow(1.02, player.constitutiveSingularityCount - 100)
```
KhafraDev, line 2649:
```suggestion
    const extraMult = constitutiveSingularityCount > 100
      ? Math.pow(1.02, constitutiveSingularityCount - 100)
```

KhafraDev (changes requested):
the player object doesn't contain `constitutiveSingularityCount`, I believe this is what you're looking for?

src/singularity.ts
ZanderHill: Yep, you're right; thanks!

PR #645, opened 2025-02-06
Campaigns

KhafraDev:


src/Campaign.ts
```diff
@@ -0,0 +1,165 @@
...
+  campaignCorruptions: Partial<Corruptions>
+  campaignModifiers: CampaignModifiers
+  limit: number
+  isMeta: boolean
+}
+
+export class CampaignManager {
+  private totalCampaignTokens: number
+  private currentCampaign: Campaign | undefined
+  private campaigns!: Record<CampaignKeys, Campaign>
```
KhafraDev, line 29: never use `private` in typescript

src/Campaign.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  constructor (campaignManagerData?: ICampaignManagerData) {
+    this.campaigns = {} as Record<CampaignKeys, Campaign>
```
KhafraDev, line 32: casting shouldn't be needed. Both branches use mostly the same logic. Also you can use Object.entries()

src/Campaign.ts
```diff
@@ -0,0 +1,165 @@
...
+      for (const campaignKey of Object.keys(campaignDatas)) {
+        const key = campaignKey as keyof typeof campaignDatas
+        this.campaigns[key] = new Campaign(campaignDatas[key], key, 0)
+      }
+      this.currentCampaign = undefined
+    }
+    this.totalCampaignTokens = this.computeTotalCampaignTokens()
+  }
+
+  computeTotalCampaignTokens = () => {
```
KhafraDev, line 56:
```suggestion
  computeTotalCampaignTokens () {
```

src/Campaign.ts
```diff
@@ -0,0 +1,165 @@
...
+  // Stored as variable out of scope
+  name: string
+  description: string
+  campaignCorruptions: CampaignLoadout
+  campaignModifiers: CampaignModifiers
+  limit: number
+  isMeta: boolean
+
+  // Saved as a variable
+  _c10Completions = 0
```
KhafraDev, line 91: 1. the comment is meaningless. It won't make sense to anyone within a month.
2. why the underscore property name? Use a [private property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties)

src/Corruptions.ts
```diff
@@ -1,12 +1,348 @@
...
+    illiteracy: 0,
+    recession: 0,
+    dilation: 0,
+    hyperchallenge: 0
+  }
+  #bonusLevels = 0
+
+  constructor (p: Partial<Corruptions>) {
+    Object.assign(this.#levels, p)
+  }
```
KhafraDev, line 63: I don't there there is ever a reason to use Partial. Whatever is being passed here should be populated with empty values instead.

src/Corruptions.ts
```diff
@@ -131,208 +439,233 @@ export const corruptionDisplay = (index: number) => {
...
-        return false
-      }
-    }
+  console.log(corruptions)
+
+  if (corruptions.includes('/') && corruptions.split('/').length === 8) {
+    corr = convertInputToCorruption(corruptions.split('/').map(Number))
+  } else {
+    corr = JSON.parse(corruptions) as Corruptions
+  }
```
KhafraDev, line 654: this should use zod

src/Corruptions.ts
```diff
@@ -131,208 +439,233 @@ export const corruptionDisplay = (index: number) => {
...
   }
-
   return false
 }
 
 async function importCorruptionsPrompt () {
   const input = await Prompt(i18next.t('corruptions.importCorruptionsPrompt.import'))
 
-  if (!applyCorruptions(`0/0/${input}/0/0/0`)) {
+  if (!applyCorruptions(input as string)) {
```
KhafraDev, line 668:
```suggestion
  if (input === null || !applyCorruptions(input)) {
```

src/Corruptions.ts
```diff
@@ -352,29 +685,28 @@ async function corruptionLoadoutGetNewName (loadout = 0) {
...
+      return Alert(i18next.t('corruptions.loadoutPrompt.errors.crazyJoke'))
     }
   }
 }
 
 export const updateCorruptionLoadoutNames = () => {
   const rows = getElementById<HTMLTableElement>('corruptionLoadoutTable').rows
-  const totalSlots = 8 + PCoinUpgradeEffects.CORRUPTION_LOADOUT_SLOT_QOL
-  for (let i = 0; i < totalSlots; i++) {
+  for (let i = 0; i < 8; i++) {
```
KhafraDev, line 698: this seems to be undoing the PseudoCoin perk?

src/History.ts
```diff
@@ -522,21 +540,28 @@ const resetHistoryFormatCorruptions = (data: ResetHistoryEntryAscend): [string,
...
-      }" title="${resetHistoryCorruptionTitles[i]}">${data.usedCorruptions[corruptionIdx]}`
-    } else {
-      corruptions += `<span>${corrs > 0 ? '/0' : '0'}</span>`
+        resetHistoryCorruptionInfo[corrKey].image
+      }" title="${resetHistoryCorruptionInfo[corrKey].title}">${corrToLoad[corrKey]}`
     }
     corrs++
   }
+
+  const corr_str = JSON.stringify(corrToLoad)
```
KhafraDev, line 561: we're in js, camelCase for consistency

src/Reset.ts
```diff
@@ -299,7 +299,7 @@ const resetAddHistoryEntry = (input: resetNames, from = 'unknown') => {
         seconds: player.ascensionCounter,
         date: Date.now(),
         c10Completions: player.challengecompletions[10],
-        usedCorruptions: player.usedCorruptions.slice(0), // shallow copy,
+        usedCorruptions: player.corruptions.used, // shallow copy,
```
KhafraDev, line 302:
```suggestion
        usedCorruptions: player.corruptions.used,
```
the comment no longer applies

src/Synergism.ts
```diff
@@ -880,7 +880,32 @@ export const player: Player = {
     6: false
   },
 
-  prototypeCorruptions: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+  corruptions: {
+    next: new CorruptionLoadout({}),
+    used: new CorruptionLoadout({}),
```
KhafraDev, line 885: it'd probably be better to allow the parameter to be optional

src/UpdateVisuals.ts
```diff
@@ -1866,3 +1865,16 @@ export const visualUpdateEvent = () => {
...
+// Never again? - 2025 Platonic
+let cachedVisualCampaignUpdates = false
+
+export const visualUpdateCampaign = () => {
+  if (!cachedVisualCampaignUpdates) {
+    DOMCacheGetOrSet('campaignsIntroduction').innerHTML = i18next.t('campaigns.intro')
+    DOMCacheGetOrSet('campaignsInfo1').innerHTML = i18next.t('campaigns.campaignInfo')
+    cachedVisualCampaignUpdates = true
+  }
+}
```
KhafraDev, line 1880: you can use the `memoize` function from Utility.ts

src/Utility.ts
```diff
@@ -231,3 +231,7 @@ export function memoize<Args extends unknown[], Ret> (fn: (...args: Args) => Ret
     return ret
   }
 }
+
+export const validateNonnegativeInteger = (n: number | string): boolean => {
+  return Number.isFinite(n) && !Number.isNaN(n) && Number.isInteger(n)
+}
```
KhafraDev, line 237: we should probably just use zod for this.

z.number().int().safe().safeParse(/*  value */).success

src/Variables.ts
```diff
@@ -449,7 +449,7 @@ export const Globals: GlobalVariables = {
   autoTalismanTimer: 0,
 
   autoChallengeTimerIncrement: 0,
-  corruptionTrigger: 1,
+  corruptionTrigger: 'illiteracy',
```
KhafraDev, line 452: possible to use an enum here?

src/saves/PlayerJsonSchema.ts
```diff
@@ -1,6 +1,20 @@
 import { z } from 'zod'
 import type { Player } from '../types/Synergism'
 import { playerSchema } from './PlayerSchema'
+import type { CorruptionLoadout, Corruptions } from '../Corruptions'
+
+const convertArrayToCorruption = (array: number[]): Corruptions => {
```
KhafraDev, line 6: I'm pretty sure this is duplicated

src/saves/PlayerJsonSchema.ts
```diff
@@ -10,6 +24,23 @@ export const playerJsonSchema = playerSchema.extend({
   wowHypercubes: z.any().transform((hypercubes: Player['wowHypercubes']) => Number(hypercubes)),
   wowPlatonicCubes: z.any().transform((cubes: Player['wowPlatonicCubes']) => Number(cubes)),
 
+  campaigns: z.any().transform((campaignManager: Player['campaigns']) => {
```
KhafraDev, line 27: z.any() shouldn't be used in the same way `any` shouldn't be used

src/saves/PlayerJsonSchema.ts
```diff
@@ -10,6 +24,23 @@ export const playerJsonSchema = playerSchema.extend({
...
+    return campaignManager.c10Completions
+  }),
+
+  corruptions: z.any().transform((stuff: Player['corruptions']) => {
```
KhafraDev, line 31: z.any() shouldn't be used in the same way `any` shouldn't be used

src/saves/PlayerJsonSchema.ts
```diff
@@ -72,4 +103,39 @@ export const playerJsonSchema = playerSchema.extend({
...
+      {} as Record<string, Corruptions>
+    )
+
+    player.corruptions.saves = corruptionSaveStuff
+  }
+
+  player.usedCorruptions = undefined
+  player.prototypeCorruptions = undefined
+  player.corruptionLoadoutNames = undefined
+  player.corruptionLoadouts = undefined
```
KhafraDev, line 138: delete-ing or Reflect.deleteProperty would be more accurate to the intent

src/saves/PlayerUpdateVarSchema.ts
```diff
@@ -0,0 +1,50 @@
+import { CorruptionLoadout, type Corruptions, CorruptionSaves } from '../Corruptions'
+import { playerSchema } from './PlayerSchema'
+
+const convertArrayToCorruption = (array: number[]): Corruptions => {
```
KhafraDev, line 4: this is definitely duplicated

src/Synergism.ts
Pseudonian: The new commit uses `corruptionsSchema.parse({})` - I tried to make the parameter optional but it kept yelling me to give it a parameter, even after using `.optional()` - after checking docs and asking Claude it seems to demand a parameter if you want to fill in default values for the object without requiring making the entire object optional or always having all params for the object populated. Will need to check into this later.

src/Utility.ts
Pseudonian: .safe() restricts integers to only within javascript Floating Point Precision, but ideally this would be used to allow any integer up to ~1e308. However this seems to not be used in this case, so I may go ahead with this suggestion anyway

src/saves/PlayerJsonSchema.ts
Pseudonian: There's some weird differences in indexing between the two (one is shifted by 2, for reasons I can't explain). The other one is definitely a duplicate though so I removed it in commit

src/Variables.ts
Pseudonian: This feeds into `corruptionTrigger` which uses `keyof Corruptions` as argument as of this update

src/saves/PlayerJsonSchema.ts
KhafraDev: please leave comment

PR #668, opened 2025-03-07
Prevent level 15 corruptions from applying in exalt1/7

galencolin:
screenshot of a [save](https://github.com/user-attachments/files/19122085/Synergism-1x30trial2-2025-03-07.01_30_56.txt) in exalt1x30 (next mult should be (base) 3 + (bonus) .99, and should say 1 free level... was so used to the display being wrong pre-campaigns that I didn't question it)
image

ig it also has the same effect on exalt7

### DIFF
--- a/src/Corruptions.ts
+++ b/src/Corruptions.ts
@@ -229,7 +229,7 @@ export class CorruptionLoadout {
   }
 
   get bonusLevels () {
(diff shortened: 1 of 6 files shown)


Pseudonian:
Good catch!

PR #675, opened 2025-03-11
Campaign progress bars

Pseudonian:


Pseudonian:
Thank you for the suggestion, gwabo!

PR #676, opened 2025-03-11
fixed challenges and corruptions

httpsnet:
- Set corruptions level to 0 if it is below the requirement on load.
For years, there has been a bug where ascending or corrupting just before a singularity could leave corruptions behind after the singularity.
Since Platonic removed the code for this in a campaign commit, more and more people have complained about this rare issue.
I've once again introduced code to fix problematic saves and make them playable.
Please do not remove the code even if you have fixed this issue, as saves and files from players who played the affected version will not be lost.

- Prevented ascension challenges from starting unless the requirement is actually met.
This is to prevent abuse of CSS styles.

- Once C10 is complete, C11 will unlock without ascending.
You will ascend when you start C11. I've tested this a few times over the past 2 years so it should be fine.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1675,7 +1675,7 @@
                     <img id="challenge10" alt="Challenge 10" class="challenge" src="Pictures/Default/Challenge10.png" loading="lazy">
                     <p class="challengeLevels" id="challenge10level">0 / 25</p>
                 </div>
-                <div class="ascendunlock">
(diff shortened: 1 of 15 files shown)


Pseudonian:
Good PR - succinct but also fixes the existing issues with Corruptions on already broken files, plus a few bug fixes.

PR #678, opened 2025-03-11
Enable tab wrapping on narrow screen such as phone

ProditorMagnus:
I am concerned why there is additionally https://github.com/Pseudo-Corp/SynergismOfficial/blob/22230e9be0e029962b8a97e159d5fb83d5b442fe/Synergism.css#L590

Example on phone size:
image
image


httpsnet:
You should remove this.style.cssText and use only CSS instead

Pseudonian:
Great!

PR #708, opened 2025-06-22
Fast Forward Hardcap into Softcap

Mixelz:
# Fast Forward Hardcap into Softcap
This PR implements a simple change:
Fast Forward is no longer hardcapped at Sing 200 and instead its effect will be reduced by 1 for every 15 Sings past it, up until the effect is entirely negated by Sing 245.

Needing to do at least 100 manual sings is a chore at best and monotonously dull at worst, this change is intended to make that climb smoother up until Sing Penalties start becoming a real problem (Roughly Sing 250).

The code was also refreshed by placing the Exalt check at the front (No doing needless maths), unrolling maths in line with other examples and commenting everything up nice and neatly.

Based on the fact we have had more content and buffs compared to when S200 - 250 was the endgame, I do not believe further changes will be warrented, however this would still be up to Plats descretion. Technically this solution is also flexible in case Plat ever decides to add new FF upgrades so that's something?

## TLDR
- Fast Forward now is a scaling softcap starting at S200, reducing by -1 for every 15 Sings completed.
- Exalt check at the front so you don't do maths on a value that gets immediately zeroed out.
- Cleaned Code around the affected area, including comments.

### DIFF
--- a/src/singularity.ts
+++ b/src/singularity.ts
@@ -2552,7 +2552,7 @@ const handlePerks = (singularityCount: number) => {
 // Indicates the number of extra Singularity count gained on Singularity reset
 export const getFastForwardTotalMultiplier = (): number => {
   let fastForward = 0
-  var ffHardCap = 245 // True Hardcap for FF (Change as needed).
+  const ffHardCap = 245 // True Hardcap for FF (Change as needed).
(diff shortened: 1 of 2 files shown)


Mixelz:
Closing this PR as it is no longer relevant! (Hardcap was completely removed!) \o/

PR #709, opened 2025-06-23
fix Cost and Effect of maxed-out PseudoCoin upgrades

pinterior:
Cost and effect for maxed-out PseudoCoin upgrades are not updated properly.
This PR fixes the problem.

image


KhafraDev:
thank you, fixed in 85bed324bd8a379536e5d166f9ec32d163ad07d2

PR #720, opened 2025-09-22
Fix Tesseract/Hypercube effects past softcap

Semenar:
In the recent update, the Tesseract/Hypercube effect formulas past softcap are missing the base of 1 (like the pre-softcap ones), which makes them weaker (by 100% as displayed). This seems to be the reason for the existence of a new timewall past Beta.


Pseudonian:
Thank you for the PR! This indeed fixes the bug you've noted.

PR #726, opened 2025-09-26
Update Ambrosia Luck and Red Luck achievement level effect to match description

Fillyosopher:
Bug Report here: https://discord.com/channels/677271830838640680/1420457591884218368/1420938026468507700

Changes the Ambrosia Luck and Green Ambrosia Luck bonus from Achievement Level Rewards to the value listed in the description of the reward.  No images included because I don't have a save that deep into the game to test with.


Pseudonian:
oh, It's actually the description that is incorrect; the effect is correct since these perks unlock at the same Synergism Level as `minLevel` which is 230 and 260 respectively.

PR #807, opened 2026-04-01
Fix some things in 4.2.0

The1AndOnly6030:
I, uh, don't really know how to start with these PRs. Ah well. Here's another long list of things I fixed/changed:
* Added red borders to Offering/Obtainium Auto to indicate that they are indeed non-refundable
* EDIT: I forgot to mention this yesterday but I also fixed Obtainium Auto not making Research cheaper
* Added Primal Power, Welcome to Singularity, and Automatic Sweep-o-Matic to Pictures/Default (I did this so the console would stop spitting errors about them not being found there, but then I realized that issue doesn't appear on live versions. Oops. :derpsmith:)
* Adjusted EXALT 7's (limited time) time limit for completions below 10 (-60 per completion -> -55 per completion) and fixed the unusually low time limit on 7x10. This _is_ a small nerf to the difficulty (although it's a fix for 7x10), though it also handles the issue where 7x9 and 7x10 would have the same time limit (120 seconds). Basically, it would go (starting from e7x8) 180 -> 120 -> 60 -> 115 -> 110, but now it goes 215 -> 160 -> 120 -> 115 -> 110. If I did the math right, that is.
* Fixed(?) Token Inheritance not actually giving any tokens at level 1 (it's a very questionable fix, but it seems to work)
* Added some coloring to stuff in stats for nerds (magenta for one/half mind, lime for events, green for synergism level, red for penalties, cyan for things only active in certain conditions. Basically just following existing patterns)
* Also adjusted the EXALT numbers for synergism stage
* The free challenge completions from "Not So Challenging" no longer count towards effective C13 completions
* Overall, stats in Stats for Nerds are described in a much more consistent manner (e.g. Quark Upgrades are almost always prefixed with "Quarks - ")
* Hopefully this time I actually removed all mentions of "Quark Shop" (except for on the Quarks tab itself)
* The penalty spike at s231 is now actually listed as s231 (not 230), and the penaltySmooth message accounts for singularity penalty reductions (originally it would appear at s270 regardless of singularity reducers)
* (Applies only to fresh saves) The buy amount for Blessings/Spirits defaults to 10 (I did this to hopefully eliminate the issue where people seemingly can't level spirits despite having the offerings for it)
* The positive salvage boost from the Achievement Talisman no longer just displays as 0% for lower rarities


The1AndOnly6030:
More new stuff:
* Fixed a message not appearing properly when you rename a corruption loadout to "crazy"
* Research costs are now actually rounded up instead of just appearing with a minimum cost of 1 (this also fixes a bug that dumb me made earlier, where maxed research claimed to cost 1 obtainium)
* In typical me fashion, more things are rewritten slightly for clarity/consistency

The1AndOnly6030:
A few more things, though 99.9% of it is just text:
* "Berries that are blue!" (oct upgrade) is now marked as QoL (this is the one non-text change. Why wasn't this one qol in the first place anyway?)
* Some i18n stuff in [red] Ambrosia no longer reuses things from Octeracts
* EXALT 7 (limited time) now mentions that Platonic DELTA is disabled in it
* If an upgrade only boosts 3-7D cubes but isn't post-Singularity content, there's a subtle indicator of that. Something like Platonic OMEGA will now say "...x1.01 'all' Cube gain per C9 completion..." Furthermore, any upgrade that truly boosts all cubes will say so. Something like Cube Flame will now say "Permanently gain x% more Cubes of all types."
* Instances of "Golden Quark Upgrade" in stats for nerds have generally been shortened to "Golden Quarks" to make it fit more with the other sets of upgrades (quarks, octeracts, ambrosia)
* Every Immaculate Obtainium multiplier (that I found, at least...) now mentions that they are indeed immaculate multipliers.
* Once again, entirely too many miscellaneous changes in en.json for me to feel like mentioning (more clarity/consistency)

The1AndOnly6030:
Since this PR was last updated, a lot of its fixes have been done in different ways or just separately from the PR itself. There's also an overwhelming amount of changes that would be much better as single PRs, so I'll just close this one.

PR #810, opened 2026-04-08
Fix a bug for Cx14 when I opened less than 40000 Platonic Cubes at once

loader3229:
Fix a bug for Cx14 when I opened less than 40000 Platonic Cubes at once
When I opened less than 40000 Platonic Cubes at once with Cx14, cube upgrade Cx14 does not apply for statue calculations. This PR will fix the bug.
To reproduce the bug: start a new singularity with Cx14 unlocked. Don't open any Platonic Cubes before Cx14 bought. After Cx14 bought, open 39999 Platonic Cubes at once.


Pseudonian:
I think there is a cleaner way of implementing this, as right now we are checking for a precise weight value (===1) as well as making the same type assertions many times in a row.

PR #812, opened 2026-04-09
Fix early potion usage

icydvorak:


Pseudonian:
Why not just use `player.highestSinguarityCount === 0`?

PR #815, opened 2026-04-13
fix scientific notation

pinterior:
Fix last digit of scientific notation.

### Before

image
image

### After

image
image


Pseudonian:
erm 
Screenshot 2026-04-17 at 7 10 35 PM

PR #826, opened 2026-05-02
Format fix

rus9384:
1. Fixed incorrect last digit display in Pure Scientific notation.
2. Long format now applies to Pure Scientific notation.
3. Fixed a bug with incorrect display of 1e-100 number in Pure Scientific notation.
4. Pure Engineering notation now displays numbers smaller than 0.001 in engineering notation.
5. Pure Engineering notation now shows the same number of significant digits regardless of exponent.
6. Pure Engineering notation now behaves as Default if not all digits of the exponent are visible.
7. Default and Pure Engineering notations now truncate the number just as Pure Scientific notation.
8. Accuracy now is correctly applied to large numbers.


src/Synergism.ts
```diff
@@ -2262,58 +2199,42 @@ export const format = (
...
-      mantissa = Math.ceil(mantissa)
-    }
-
-    return (
-      Math.floor(mantissa * Math.pow(10, accuracy)) / Math.pow(10, accuracy)
-    ).toLocaleString(undefined, {
-      minimumFractionDigits: accuracy,
-      maximumFractionDigits: accuracy
-    })
+    return input.toLocaleString(undefined, formatOpts).toLowerCase() // We want 'e' to be in lower case
```
KhafraDev: I would inline formatOpts here

src/Synergism.ts
```diff
@@ -2360,50 +2281,61 @@ export const format = (
...
-    if ((power < 2 || (long && power < 3)) && accuracy > 0) {
-      standardString = standard.toFixed(
-        power === 2 && accuracy > 2 ? 2 : accuracy
-      )
-    } else {
-      // If it doesn't fit those criteria drop the decimal places
-      standard = Math.floor(standard)
-      standardString = standard.toString()
+
+    const formatOpts = {
```
KhafraDev: same here, formatOpts should be inlined

src/Synergism.ts
```diff
@@ -2360,50 +2281,61 @@ export const format = (
...
+    power -= powerOver
+    mantissa *= Math.pow(10, powerOver)
+  }
+
+  // Number.toLocaleString opts for 'accuracy' decimal places
+  const locOpts = { 
+    minimumSignificantDigits: accuracy,
+    maximumSignificantDigits: accuracy,
+    roundingMode: 'trunc' as const
+  }
```
KhafraDev, line 2323:
```suggestion
  const locOpts: Intl.NumberFormatOptions = { 
    minimumSignificantDigits: accuracy,
    maximumSignificantDigits: accuracy,
    roundingMode: 'trunc'
  }
```

we should use the types TypeScript provides

src/Synergism.ts
rus9384: Changed

src/Synergism.ts
rus9384: Changed

src/Synergism.ts
rus9384: Done

PR #827, opened 2026-05-24
Update singularity.ts

rus9384:
Fixed Salvage and ELO penalty at low Singularities, if you have Singularity reductions.


Pseudonian:
Good catch!

PR #843, opened 2026-07-05
Buy fix

rus9384:
Bug Fixes:
1. Fixed Building Cost formula.
2. Fixed Diamond and Particle Buildings cost on reset when reset starts with 1 unit of that building.
3. Fixed Particle Building manual buy amount (it was buying 1 more, so 2 instead of 1, 11 instead of 10, etc.).

Other:
1. Made manual Building buy use the same function as auto-buy.
2. Added code support for 'max' amount for manual buy (this does *not* add UI support for that).


Pseudonian:
Thank you for the PR! Looks solid.

PR #844, opened 2026-07-12
Fix tab row overflow on narrow screens

RYGHT1:
Fixes the tab row scrollbar flickering on smaller screens or especially during editing tab visibility and tones down the mobile bounce animation.

I assumed the mobile hamburger menu removed the necessity of the scrollbar being there in the first place.
synergism scrollbar flicker


KhafraDev:
thank you!

PR #845, opened 2026-07-12
Feat/toggle tabs

RYGHT1:
makes tab removal mode a toggle so hidden tabs can be seen and restored while editing.
tab-toggle


KhafraDev:
thanks again!

PR #846, opened 2026-07-14
Format fix

rus9384:
Fixes:
1. Fixed displayed value in RNG-Based Obtainium Booster and RNG-Based Offering Booster Ambrosia Upgrades.
2. Fixed incorrect rounding for numbers, in particular, for percent values.
3. Fixed Auto-Transcension localization.
4. Fixed `-0` Talisman Fragment amount.
5. Fixed Tributum Formicidae's effect sign.
6. Fixed Ant Auto-Sacrifice Immortal ELO threshold when it's set to a non-integer value.

Other Changes:
1. Removed long exponent number format (redundant and not improving readability).
2. Removed fractional number format (for being confusing).
3. Added truncated number format.
4. Added number locale support in English localization (via a new i18n plugin).
5. Removed long notation for Talisman Fragments (it was not fitting in the element).
6. Removed long notation for Abyss Hepteract cost of Platonic Upgrades (for consistency).
7. Removed long notation of Hepteracts.
8. Added formatting for opened Hepteract percentage.
9. Increased accuracy for daily levels of Octeract Upgrades.
10. Increased accuracy of some other numbers (Rune and Talisman effects, Hepteract effects, etc.).

Find & replace steps applied to `en.json` file:
1. Enable regex mode
2. Find: `(\d{1,3}((\.\d+)?e\d+(,?\d{3})*|(,\d{3})+(\.\d+)?|\.\d+))` => Replace: `num:(($1))`
3. Find: `(num:\(\((?:\d|e)*\d),(\d[^)]*\)\))` => Replace: `$1$2` (repeat until no matches)

### DIFF
--- a/src/Plugins/NumberLocale.ts
+++ b/src/Plugins/NumberLocale.ts
@@ -1,12 +1,13 @@
-import type { Module } from 'i18next'
+import type { PostProcessorModule } from 'i18next'
 import { format } from '../Synergism'
 import Decimal from 'break_infinity.js'
 
 export default {
   type: 'postProcessor',
   name: 'NumberLocale',
-  process: (value: string): string => value.replace(
-    /num:\(\(([^e)]*(e?)[^e)]*)\)\)/g, //num:1e6 => 1e6; num:1e6L => 1,000,000; num:0.1% => 0.1%
+  process: (value: string): string => value.includes('num:') ? value.replace(
+    // Regex replacement is faster than doing it in a loop
+    /num:\(\(([^e)]*(e?)[^)]*)\)\)/g, //num:1e6000 => 1e6,000; num:1000000 => 1,000,000; num:0.1% => 0.1%
     (_, num, exp) => format(Decimal.fromString(num), 5, exp !== 'e', true)
-  )
-} as Module
+  ) : value
+} satisfies PostProcessorModule
--- a/src/Plugins/NumberLocale.ts
+++ b/src/Plugins/NumberLocale.ts
@@ -5,9 +5,17 @@ import Decimal from 'break_infinity.js'
 export default {
   type: 'postProcessor',
   name: 'NumberLocale',
-  process: (value: string): string => value.includes('num:') ? value.replace(
-    // Regex replacement is faster than doing it in a loop
-    /num:\(\(([^e)]*(e?)[^)]*)\)\)/g, //num:1e6000 => 1e6,000; num:1000000 => 1,000,000; num:0.1% => 0.1%
-    (_, num, exp) => format(Decimal.fromString(num), 5, exp !== 'e', true)
-  ) : value
+  process: (value: string): string => {
+    let result = ''
+    let start = 0, index = 0
+    while ((index = value.indexOf('num:((', start)) !== -1) {
+      const end = value.indexOf('))', index)
+      const num = value.slice(index + 6, end)
+      result += value.slice(start, index)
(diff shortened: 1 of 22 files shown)


src/Plugins/NumberLocale.ts
```diff
@@ -0,0 +1,12 @@
+import type { Module } from 'i18next'
+import { format } from '../Synergism'
+import Decimal from 'break_infinity.js'
+
+export default {
+  type: 'postProcessor',
+  name: 'NumberLocale',
+  process: (value: string): string => value.replace(
```
KhafraDev: if (value.contains('num:') before the regex

for performance reasons

src/Plugins/NumberLocale.ts
```diff
@@ -0,0 +1,12 @@
...
+    /num:\(\(([^e)]*(e?)[^e)]*)\)\)/g, //num:1e6 => 1e6; num:1e6L => 1,000,000; num:0.1% => 0.1%
+    (_, num, exp) => format(Decimal.fromString(num), 5, exp !== 'e', true)
+  )
+} as Module
```
KhafraDev: satisfies Module?

src/Synergism.ts
```diff
@@ -2176,26 +2174,28 @@ export const format = (
...
-    if (input === 0) {
-      return '0'
-    }
-
-    // Gets power and mantissa if input is of type number and isn't 0
-    power = Math.floor(Math.log10(input as number))
-    mantissa = (input as number) / Math.pow(10, power)
-  } else if (input instanceof Decimal) {
+  let power: number = 0
+  let mantissa: number = 0
```
KhafraDev:
```suggestion
  let power = 0
  let mantissa = 0
```

src/Synergism.ts
```diff
@@ -2176,26 +2174,28 @@ export const format = (
...
+  if (input instanceof Decimal) {
```
KhafraDev:
```suggestion
  if (inputType === 'object' && input instanceof Decimal) {
```
instanceof should always be avoided

src/Plugins/NumberLocale.ts
KhafraDev: it should be relatively easy avoiding the regex entirely

KhafraDev (dismissed):

src/Plugins/NumberLocale.ts
rus9384: Added 'includes' check

src/Plugins/NumberLocale.ts
rus9384: Changed to `satisfies PostProcessorModule`

src/Synergism.ts
rus9384: Done

src/Synergism.ts
rus9384: Reverted type check order, added a mistype check.

src/Plugins/NumberLocale.ts
KhafraDev: the regex should be replaced. Do not trust microbenchmarks

KhafraDev: https://benediktmeurer.de/2016/12/16/the-truth-about-traditional-javascript-benchmarks/

rus9384: Replaced regex with loop scanning

Pseudonian:
Thank you for the PR. Looks great!

PR #854, opened 2026-08-07
Buy fix

rus9384:
1. Fixed Accelerator Boost formule before Mythos Upgrade 2x1.
2. Added Accelerator Boost cost update on buying Thrift Blessing.


Pseudonian:
There are potential long-term issues with relying on the `player` object to store costs. While this PR does address bugs related to buying Accelerator Boosts, the fix suggests that dynamic cost adjustments can pose an issue for building costs when storing the values in `player`.

rus9384:
While I agree that in general some larger overhaul is needed, the main fix is a misaligned Accelerator Boost cost index in the part of the code that handles Accelerator Boost purchases before Mythos Upgrade 2x1. Due to which you basically get one more Accelerator Boost.

PR #860, opened 2026-08-14
Roomba flickering fix

rus9384:
Fixed roomba flickering

### DIFF
--- a/src/Research.ts
+++ b/src/Research.ts
@@ -311,16 +311,23 @@ export const updateResearchAuto = (index: number) => {
 }
 
 export const advanceResearchRoomba = () => {
+  let maxIndex = player.autoResearch
   if (isResearchMaxed(player.autoResearch) || !isResearchUnlocked(player.autoResearch)) {
-    player.roombaResearchIndex = Math.min(researchOrderByCost.length - 1, player.roombaResearchIndex + 1)
-    player.autoResearch = researchOrderByCost[player.roombaResearchIndex]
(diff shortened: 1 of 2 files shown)


KhafraDev:
@claude review this PR

rus9384:
The only quirk of using `player.autoResearch` instead of `researchOrderByCost.length - 1` is that the fewer researches the player has researched, the faster the roomba is. Which still is maybe not how it should be, but it wouldn't break logic the way Claude claims, as that scenario is covered with `isResearchMaxed(player.autoResearch)`.

The modulus is also specifically related to the affordability check in the comment.

KhafraDev:
@claude The only quirk of using player.autoResearch instead of researchOrderByCost.length - 1 is that the fewer researches the player has researched, the faster the roomba is. Which still is maybe not how it should be, but it wouldn't break logic the way Claude claims, as that scenario is covered with isResearchMaxed(player.autoResearch).

The modulus is also specifically related to the affordability check in the comment.

rus9384:
I think there is indeed unintended behavior with my change, but it's not exactly what Claude claimed in the first review... it's that `player.autoResearch` is the index in the unsorted array, which can be smaller than its index in the sorted array, which in some rare scenarios could make it not fallback to a cheaper research after unlocking it. So yeah, I guess just using `maxIndex = researchOrderByCost.length - 1` is the safe solution.

KhafraDev:
@claude review the pr now

